PR #3, opened 2023-06-29
Add Cluster resource template

nprokopic:
Note: WIP, initially based on Cluster resource from cluster-aws.

### What does this PR do?

Cluster resource is deployed.

### How does it look like?

```
apiVersion: cluster.x-k8s.io/v1beta1
kind: Cluster
metadata:
  annotations:
    cluster.giantswarm.io/description: "Awesome Giant Swarm cluster"
  labels:
    app: "cluster"
    app.kubernetes.io/managed-by: "Helm"
    app.kubernetes.io/version: "0.0.0-dev"
    application.giantswarm.io/team: "turtles"
    giantswarm.io/cluster: "awesome"
    giantswarm.io/organization: "giantswarm"
    giantswarm.io/service-priority: highest
    cluster.x-k8s.io/cluster-name: "awesome"
    helm.sh/chart: "cluster-0.0.0-dev"
    cluster-apps-operator.giantswarm.io/watching: ""
  name: awesome
  namespace: default
spec:
  clusterNetwork:
    services:
      cidrBlocks:
        - 172.31.0.0/16
    pods:
      cidrBlocks:
        - 100.64.0.0/12
  controlPlaneRef:
    apiVersion: controlplane.cluster.x-k8s.io/v1beta1
    kind: KubeadmControlPlane
    name: awesome
  infrastructureRef:
    apiVersion: infrastructure.cluster.x-k8s.io/v1beta1
    kind: AWSCluster
    name: awesome
```

Working towards a PoC for restructured cluster apps (more details [here](https://github.com/giantswarm/giantswarm/issues/27020)).

PTAL :)

Not yet.


nprokopic:
It's expected that JSON schema validation is failing here, because the current validation has rules for existing cluster apps, and those rules do not work in the cluster chart.

helm/cluster/templates/cluster-api/cluster.yaml
```diff
@@ -0,0 +1,36 @@
...
+    {{- if .Values.metadata.labels }}
+    {{- range $key, $val := .Values.metadata.labels }}
+    {{ $key }}: {{ $val | quote }}
+    {{- end }}
+    {{- end }}
+  name: {{ include "resource.default.name" $ }}
+  namespace: {{ .Release.Namespace }}
+spec:
+  clusterNetwork:
+    services:
```
njuettner, line 20: Do we wanna sort it similar to the CR's?
That would mean starting with pods and then services in `clusterNetwork`

helm/cluster/templates/cluster-api/cluster.yaml
```diff
@@ -0,0 +1,36 @@
...
+      cidrBlocks:
+      {{- toYaml .Values.connectivity.network.services.cidrBlocks | nindent 8 }}
+    pods:
+      cidrBlocks:
+      {{- toYaml .Values.connectivity.network.pods.cidrBlocks | nindent 8 }}
+  controlPlaneRef:
```
njuettner, line 26: Inside `controlPlaneRef` and `infrastructureRef` we do have the field `namespace` which we could define already

njuettner (approved):

helm/cluster/templates/cluster-api/cluster.yaml
nprokopic: Sure :+1:

helm/cluster/templates/cluster-api/cluster.yaml
nprokopic: We don't set namespace in these `*Ref` fields, as then it uses the same namespace as the resource which has the reference (`Cluster` here).

I think there is/was even a discussion in upstream CAPI to change the reference type to something like `LocalReference` so you cannot even set the namespace even if you wanted.

helm/cluster/values.yaml
```diff
@@ -1,3 +1,27 @@
...
-{}
+connectivity:
+  network:
+    pods:
+      cidrBlocks:
+        - 100.64.0.0/12
+    services:
+      cidrBlocks:
+        - 172.31.0.0/16
+internal:
```
nprokopic, line 11: We would statically set these from the `cluster-$provider` app in its Helm values like this:
```
cluster:
  internal:
    ...
```

and these would be treated as an internal API, as in internal in our apps, where the `cluster` chart exposes it to the `cluster-$provider` , but it's not meant to be set by end users. This is roughly aligned with the current behavior of `Values.internal`.

helm/cluster/templates/cluster-api/cluster.yaml
njuettner: Agree

helm/cluster/templates/cluster-api/cluster.yaml
njuettner: I see that make sense

PR #16, opened 2023-10-18
Normalize schema and disable cluster-app rule-set

nprokopic:
### What does this PR do?

The PR normalizes the schema, as until this point it was built without being normalized (fast-paced development, it was changing a lot, and I forgot to run normalization it the whole process).

The PR also disables cluster-app rule-set in schema lint, as that is still not working for the `cluster` chart since it has some assumptions that should be updated (like which top level properties we must have in the schema). We will enable this again when we update the schema linter.

### DIFF
--- a/.github/workflows/zz_generated.json_schema_validation.yaml
+++ b/.github/workflows/zz_generated.json_schema_validation.yaml
@@ -30,8 +30,6 @@ jobs:
       - name: Run schemalint
         id: run-schemalint
         uses: giantswarm/schemalint/actions/verify-helm-schema@v2
-        with:
-          rule-set: 'cluster-app'
   generate:
     name: Check that values.yaml is generated from values.schema.json with helm-values-gen
     runs-on: ubuntu-latest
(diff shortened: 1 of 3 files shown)


nprokopic:
Ignore Helm rendering for now. I'll check the team label failure.

nprokopic:
CI error:
```
2023-10-19 11:27:38,821 helm INFO: './helm/cluster/templates/_helpers.tpl' must contain a line that matches the regexp 're.compile('[ \\t]*application\\.giantswarm\\.io/team:[ \\t]+{{[ \\t]*index[ \\t]+\\.Chart\\.Annotations[ \\t]+"application\\.giantswarm\\.io/team"[ \\t]*(\\|[ \\t]*default[ \\t]+\\"[a-zA-Z0-9]+\\"[ \\t]+){0,1}\)'.
```

I think it fails because we have this (notice `$` in `$.Chart`)
```
application.giantswarm.io/team: {{ index $.Chart.Annotations "application.giantswarm.io/team" | quote }}
```

instead of this (just `.Chart`)
```
application.giantswarm.io/team: {{ index .Chart.Annotations "application.giantswarm.io/team" | quote }}
```

njuettner (approved):

PR #21, opened 2023-10-21
Fix KubeadmControlPlane

nprokopic:
### What does this PR do?

This PR fixes control plane resources:
- Add missing machine template to KubeadmControlPlane.

Control plane resources are templated correctly.

### How does it look like?

Updated KubeadmControlPlane resource:

```
/spec  (KubeadmControlPlane/org-giantswarm/awesome)
  + two map entries added:
    machineTemplate:
      metadata:
        labels:
          app: cluster
          app.kubernetes.io/managed-by: Helm
          app.kubernetes.io/version: 0.1.0-dev
          application.giantswarm.io/team: turtles
          giantswarm.io/cluster: awesome
          giantswarm.io/organization: giantswarm
          giantswarm.io/service-priority: highest
          cluster.x-k8s.io/cluster-name: awesome
          helm.sh/chart: cluster-0.1.0-dev
          another-cluster-label: label-2
          some-cluster-label: label-1
      infrastructureRef:
        name: awesome-control-plane-f786309f
        apiVersion: infrastructure.cluster.x-k8s.io/v1beta1
        kind: AWSMachineTemplate
    replicas: 3
```

https://github.com/giantswarm/roadmap/issues/2742

### DIFF
--- a/helm/cluster/templates/clusterapi/controlplane/_helpers.tpl
+++ b/helm/cluster/templates/clusterapi/controlplane/_helpers.tpl
@@ -1,0 +1,8 @@
+{{- define "cluster.internal.test.controlPlane.machineTemplate.spec" -}}
+template:
+  metadata:
+    cluster.x-k8s.io/role: controlPlane
+    {{- include "cluster.labels.common" $ | nindent 4 }}
+  spec:
+    machineType: "superBig"
+{{- end -}}
--- a/helm/cluster/templates/clusterapi/controlplane/kubeadmcontrolplane.yaml
+++ b/helm/cluster/templates/clusterapi/controlplane/kubeadmcontrolplane.yaml
@@ -11,15 +11,19 @@ metadata:
     {{- end }}
   labels:
     {{- include "cluster.labels.common" $ | nindent 4 }}
-    {{- if $.Values.global.metadata.labels }}
-    {{- range $key, $val := $.Values.global.metadata.labels }}
-    {{ $key }}: {{ $val | quote }}
-    {{- end }}
-    {{- end }}
+    {{- include "cluster.labels.custom" $ | indent 4 }}
   name: {{ include "cluster.resource.name" $ }}
   namespace: {{ .Release.Namespace }}
 spec:
-  version: v{{ trimPrefix "v" .Values.internal.kubernetesVersion }}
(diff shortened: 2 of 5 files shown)


helm/cluster/templates/clusterapi/controlplane/_helpers.tpl
```diff
@@ -0,0 +1,8 @@
+{{- define "cluster.internal.test.controlPlane.machineTemplate.spec" -}}
```
njuettner, line 1: Is the `test` intended?

helm/cluster/templates/clusterapi/controlplane/kubeadmcontrolplane.yaml
```diff
@@ -51,4 +55,6 @@ spec:
     {{- include "cluster.internal.controlPlane.kubeadm.postKubeadmCommands" $ | indent 4 }}
     users:
     {{- include "cluster.internal.kubeadm.users" $ | indent 4 }}
+  replicas: {{ .Values.global.controlPlane.replicas }}
+  version: v{{ trimPrefix "v" .Values.internal.kubernetesVersion }}
```
njuettner, line 59: why do we have to trim the prefix `v` and then adding it again?

njuettner (approved):

helm/cluster/templates/clusterapi/controlplane/_helpers.tpl
nprokopic: Yes, this is just a "dummy" helper that is used in the CI.

helm/cluster/templates/clusterapi/controlplane/kubeadmcontrolplane.yaml
nprokopic: I think it's in order to support `.Values.internal.kubernetesVersion` to both have and not have the `v` prefix, i.e. we don't care if the prefix is there, because we "clean up" the version from potential prefix and then we ensure it's there.

Took it from cluster-aws here https://github.com/giantswarm/cluster-aws/blob/acf1b9c712a51b364f5bece1481b3b3e3efdc263/helm/cluster-aws/templates/_control_plane.tpl#L247.

PR #23, opened 2023-10-27
Add values required for vintage migration

nprokopic:
### What does this PR do?

This PR adds missing values that configure API server, etcd and other components that need to be configured in some way during migration from vintage.

### How does it look like?

https://github.com/giantswarm/roadmap/issues/2742

https://github.com/giantswarm/cluster-aws/pull/354


helm/cluster/ci/ci-values.yaml
```diff
@@ -86,6 +86,15 @@ internal:
           - PodSecurityPolicy
           apiAudiences:
             templateName: "cluster.test.kubeadmControlPlane.kubeadmConfigSpec.clusterConfiguration.apiServer.apiAudiences"
+          etcdPrefix: giantswarm.io
```
njuettner, line 89: I'm wondering if it would make sense to move `etcdPrefix` into `etcd` object

njuettner (approved):

helm/cluster/ci/ci-values.yaml
nprokopic: It is API server CLI flag (related to etcd usage), see here https://v1-24.docs.kubernetes.io/docs/reference/command-line-tools-reference/kube-apiserver/

```
--etcd-prefix string     Default: "/registry"

The prefix to prepend to all resource paths in etcd.
```

njuettner: All good, just thought that the contract would be nicer having everything in `etcd` 🙂. Thanks 👍🏻

PR #24, opened 2023-11-09
Add Teleport support

nprokopic:
### What does this PR do?

This PR adds Teleport support that was originally added in cluster-aws here https://github.com/giantswarm/cluster-aws/pull/334.

Teleport can be optionally enabled.

### How does it look like?

`cluster` chart Helm values to enable Teleport:
```
internal:
  teleport:
    enabled: true
```

- https://github.com/giantswarm/roadmap/issues/2742
- https://github.com/giantswarm/cluster-aws/pull/334

Check if the Teleport has been ported properly from cluster-aws.

We don't yet have proper docs for the `cluster` chart.

We have yet to write proper release notes for the `cluster` chart. For now it's just about porting stuff over from `cluster-aws`, so we will add a summary before the first release.

### DIFF
--- a/Makefile.development.mk
+++ b/Makefile.development.mk
@@ -14,6 +14,4 @@ generate-schema: ensure-schema-gen ## Generate the values.schema.json file from 
 .PHONY: template
 template: ## Output the rendered Helm template
 	@cd helm/cluster && \
-		sed -i 's/version: \[/version: 1 #\[/' Chart.yaml && \
-		helm template -f ci/ci-values.yaml --debug . && \
-		sed -i 's/version: 1 #\[/version: \[/' Chart.yaml
+		helm template -f ci/ci-values.yaml --debug .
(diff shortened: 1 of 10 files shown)


nprokopic:
For some reason the above diff for Ignition shows all systemd units as removed, and new systemd units added, while the only diff is Teleport systemd unit being added. Files diff looks fine.

nprokopic:
Known issues with schema linting.

nprokopic:
`Restart=on-failure` should also be added to `audit-rules.service` systemd unit like this:

```
- name: audit-rules.service
  enabled: true
  dropins:
  - name: 10-wait-for-containerd.conf
    contents: |
      [Service]
      ExecStartPre=/bin/bash -c "while [ ! -f /etc/audit/rules.d/containerd.rules ]; do echo 'Waiting for /etc/audit/rules.d/containerd.rules to be written' && sleep 1; done"
      Restart=on-failure
```

But that will be added in the `cluster` chart's Helm values when `cluster` chart is used in `cluster-aws` because almost all systemd units will be initially specified in every cluster-<provider> app, until we figure out which are really common across all providers. (Teleport systemd unit is the first one that is included directly in the `cluster` chart, if Teleport is enabled.)

PR #26, opened 2023-12-01
Make schema compatible with cluster-aws

nprokopic:
### What does this PR do?

Updating global property names so they are the same as in cluster-aws.

### DIFF
--- a/helm/cluster/ci/ci-values.yaml
+++ b/helm/cluster/ci/ci-values.yaml
@@ -51,8 +51,8 @@ global:
       nodeLabels:
         workload-type: robots
   components:
-    cri:
-      registries:
+    containerd:
+      containerRegistries:
         docker.io:
         - endpoint: registry-1.docker.io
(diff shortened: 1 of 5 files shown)


njuettner (approved):
Could you check if we have the cluster autoscaler annotation for machine pools?

IIRC we never added into cluster only on machinepool in cluster-aws

nprokopic:
> Could you check if we have the cluster autoscaler annotation for machine pools?
> 
> IIRC we never added into cluster only on machinepool in cluster-aws

Yeah, it's not there yet. Will add that (in other PRs, not in this one) and few other recent changes we miss from cluster-aws.

nprokopic:
Also I think we can fix the CI here now, after all the work in schemalint and schemadocs.

nprokopic:
For this PR this being 🟢 is important :)

> Compare Helm Rendering / check-cmp-state (pull_request)

PR #28, opened 2023-12-06
Make schema less strict

nprokopic:
### What does this PR do?

This PR changes schema validation so it is less strict when validating OIDC and proxy. This change is done because `schemalint` currently does fully support using JSON schema with `oneOf`, `if` and `then`.

- Schema validation works in the CI.
- OIDC and proxy properties are not validated in the schema and it's easier to provide incorrect Helm values.

### How does it look like?

See files diff.


nprokopic:
> OIDC and proxy properties are not validated in the schema and it's easier to provide incorrect Helm values.

This is the case also today in `cluster-aws`, which I tried to fix in the `cluster` chart, but it didn't work.

njuettner (approved):

PR #29, opened 2023-12-06
Fix schema so we start using it in cluster-aws

nprokopic:

### What does this PR do?

This PR fixes schema and adds missing properties that are needed in order to start using the `cluster` chart in `cluster-aws` and render Cluster resource from it (only Cluster resource for now).

Changes:
- Renamed `Values.internal` to `Values.providerIntegeration`.
- Added `Values.clusterAnnotationsTemplateName` property that is used to set the name of the template that is rendering provider-specific annotations on the Cluster resource.
- Added `Values.pauseProperties` property where you can specify a list of properties and their corresponding values that will trigger setting `spec.paused` to `true` in the Cluster resource. 

`cluster` chart can be used in `cluster-aws` for rendering `Cluster` resource.

- https://github.com/giantswarm/roadmap/issues/2739
- https://github.com/giantswarm/roadmap/issues/3047

### DIFF
--- a/helm/cluster/values.schema.json
+++ b/helm/cluster/values.schema.json
@@ -1031,10 +1031,10 @@
                         }
                     }
                 },
-                "clusterLabelsTemplateName": {
+                "clusterAnnotationsTemplateName": {
                     "type": "string",
-                    "title": "Cluster labels template name",
-                    "description": "The name of the template that renders provider-specific labels for the Cluster resource"
+                    "title": "Cluster annotations template name",
+                    "description": "The name of the template that renders provider-specific annotations for the Cluster resource"
                 },
                 "components": {
                     "type": "object",
--- a/helm/cluster/values.schema.json
+++ b/helm/cluster/values.schema.json
@@ -1031,6 +1031,11 @@
                         }
                     }
                 },
+                "clusterLabelsTemplateName": {
+                    "type": "string",
+                    "title": "Cluster labels template name",
+                    "description": "The name of the template that renders provider-specific labels for the Cluster resource"
+                },
                 "components": {
                     "type": "object",
                     "title": "Components",
--- a/helm/cluster/values.schema.json
+++ b/helm/cluster/values.schema.json
@@ -604,7 +604,6 @@
                         "baseDomain",
                         "network"
                     ],
-                    "additionalProperties": false,
                     "properties": {
                         "baseDomain": {
                             "type": "string",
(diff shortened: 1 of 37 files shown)


nprokopic:
Will fix schema linting and docs shortly.

nprokopic:
Docs fixed, schema linting is still failing because it is still picking up older version when installing schemalint with 

```
go install github.com/giantswarm/schemalint/v2@v2
```

helm/cluster/values.schema.json
```diff
@@ -850,12 +844,10 @@
                     "type": "object",
                     "title": "Metadata",
                     "required": [
-                        "description",
```
njuettner, line 853: Hmm looking at cluster-was it seems description is required, also a new field is required `preventDeletion`

helm/cluster/values.schema.json
```diff
@@ -1544,37 +1534,90 @@
                             "default": true
                         }
                     },
-                    "oneOf": [
+                    "allOf": [
```
njuettner, line 1537: Maybe it's worth to add some documentation here, why we need it and what it does

helm/cluster/values.schema.json
nprokopic: I also thought that description was required, but apparently it's optional. It's not enforced [here](https://github.com/giantswarm/cluster-aws/blob/ff1934aa2684fcea788f61ad460f1ef1ee884c99/helm/cluster-aws/values.schema.json#L737), and I checked with KaaS Area this last week since I wasn't sure.

> also a new field is required preventDeletion

Damn, right, forgot that one, thanks!

helm/cluster/values.schema.json
nprokopic: Yeah, definitely, for this one and many more parts of the schema, especially these more complex parts.

njuettner (approved):

helm/cluster/values.schema.json
nprokopic: For this part here it's a bit harder to have "inline" documentation that explains if/then logic, because not all JSON schema objects can have description, so we can explain this on a higher level in cluster chart docs.

helm/cluster/values.schema.json
nprokopic: Added `preventDeletion`.

nprokopic:
schema lintting finally after fixing and releasing schemalint tool v2.5.1 💚

PR #37, opened 2023-12-21
Add Helm value for specifying the provider name

nprokopic:
And render `app: cluster-<provider>` label instead of `app: cluster` label.

### What does this PR do?

- Add Helm value for specifying the provider name.
- Render `app: cluster-<provider>` label instead of `app: cluster` label.

When deploying a WC, `app: cluster-<provider>` is unmodified.

Cluster upgrade tests are failing in https://github.com/giantswarm/cluster-aws/pull/460, because we have ClusterResourceSet that has `app: cluster-aws` in the selector.

Changing `app: cluster-aws` to `app: cluster` in Cluster resources breaks the ClusterResourceSet's selector. OTOH we cannot change ClusterResourceSet's selector because it is immutable.

So the fastest fix here, and that maybe even makes sense long-term, is to render `app: cluster-<provider>` label in the cluster chart. It may make sense even as a long-term permanent change (not just a fix for this) because `cluster-<provider>` is the app that we are actually deploying, while `cluster` is the subchart.


fiunchinho:
Wouldn't this make it weird? The CRs would contain labels like
```
    app: cluster-aws
    app.kubernetes.io/managed-by: Helm
    app.kubernetes.io/version: 0.1.0
    helm.sh/chart: cluster-0.1.0
```

The `app` label and `helm.sh/chart` wouldn't mention the same chart/app. Also `0.1.0` is not the version of what's specified in the `app` label.

If I think about the purpose of the `clusterSelector` field on the `ClusterResourceSet` CR, I believe the best would be to only select clusters based on their name (and maybe `watch-filter`, I'm not sure if that's still required).
```
spec:
  clusterSelector:
    matchLabels:
      cluster.x-k8s.io/cluster-name: t-lhqb2d2zh9dd9hn659
```

nprokopic:
> Wouldn't this make it weird? The CRs would contain labels like
>
>    app: cluster-aws
>    app.kubernetes.io/managed-by: Helm
>    app.kubernetes.io/version: 0.1.0
>    helm.sh/chart: cluster-0.1.0
> The app label and helm.sh/chart wouldn't mention the same chart/app. Also 0.1.0 is not the version of what's specified in the app label.

Yeah, well, yes and no IMO 😄

These two being different is weird because there is name of one app, but version from the other app (and only `cluster-aws` here is deployed as the app).

```
app: cluster-aws
app.kubernetes.io/version: 0.1.0
```

So here we should have both from cluster-aws like this:
```
app: "cluster-aws"
app.kubernetes.io/version: 0.50.0
```

I don't know if parent chart can pass its version to a subchart, I did not found a way (setting it in some random Helm value would be weird and error prone), so we can maybe just not set `app.kubernetes.io/version` at all.

Then, chart label being different is fine I would say, because now we really have multiple charts, and resources really come from different charts, so it could be even helpful to see which chart is rendering the resource.

```
helm.sh/chart: cluster-0.1.0
```

> If I think about the purpose of the clusterSelector field on the ClusterResourceSet CR, I believe the best would be to only select clusters based on their name

Yeah, agreed there, I would maybe add organisation name to the cluster name. Just not sure how we can change that selector in ClusterResourceSet, because the selector is immutable. Maybe by deleting ClusterResourceSet first (in a pre-install hook?) and then let Helm redeploy it with new selector labels. So from what I see, we either find a way to do this and have new simpler selector labels, or we set `app: cluster-aws` label in cluster chart so it's backward compatible.

nprokopic:
So current idea for making this better:
- We fix selector labels like described above
- We add a Helm pre-upgrade hook that deletes ClusterResourceSets, so that Helm can install ClusterResourceSets with new label selectors

And we can maybe completely ditch these two labels:
```
app: cluster-aws
app.kubernetes.io/version: 0.1.0
```

And maybe just set
```
app.kubernetes.io/part-of: cluster-aws
```

Which is recommended here https://kubernetes.io/docs/concepts/overview/working-with-objects/common-labels/#labels:

> The name of a higher level application this one is part of

nprokopic:
> and maybe watch-filter, I'm not sure if that's still required

I guess we can leave it in common labels for now, just in case, but I don't think we need it in the selector labels there.

nprokopic:
> And maybe just set
>
>     app.kubernetes.io/part-of: cluster-aws

Or set
```
app.kubernetes.io/name: cluster-aws
app.kubernetes.io/component: cluster
```

fiunchinho:
Maybe it's not worth changing `ClusterResourceSets` selector and mess with that, as they will go away once we upgrade to 1.25 and PSPs are removed.

fiunchinho (approved):

nprokopic:
> Maybe it's not worth changing `ClusterResourceSets` selector and mess with that, as they will go away once we upgrade to 1.25 and PSPs are removed.

Right, good point, so at least for now we keep `app: cluster-aws` label.

I will also remove the label `app.kubernetes.io/version` that has cluster chart version (we will still have `helm.sh/chart: cluster-0.1.0`) and remove the weirdness there.

And set these two labels:

```
app.kubernetes.io/name: cluster-aws
app.kubernetes.io/component: cluster
```

`app.kubernetes.io/component` sounds more correct than `app.kubernetes.io/part-of`.

Gacko:
Isn't there a possibility to get the labels from the parent chart?

nprokopic:
> I checked the Helm labels guidelines here https://helm.sh/docs/chart_best_practices/labels/
> 
> According to Helm guidelines it would be something like this (different than what I wrote above):
> 
> ```
>     app.kubernetes.io/name: "cluster"
>     app.kubernetes.io/version: "0.1.0"
>     app.kubernetes.io/part-of: "cluster-aws"
>     app.kubernetes.io/instance: "awesome"
>     app.kubernetes.io/managed-by: "Helm"
> ```
> 
> So all labels are about the cluster chart (the one that is rendering the resource), and `app.kubernetes.io/part-of` is the only one pointing to the parent chart.

So I would go with these Helm guidelines, as they are a bit more specific/precise than the Kubernetes "general" labels guidelines [here](https://kubernetes.io/docs/concepts/overview/working-with-objects/common-labels/#labels).

And we keep this with a comment:

```
    # app label is deprecated and it will be removed after upgrading to Kubernetes 1.25
    app: "cluster-aws"
```

Gacko:
I really like the idea of going with Helm best practices, they sound good! Do they fit our needs?

nprokopic:
> I really like the idea of going with Helm best practices, they sound good! Do they fit our needs?

Yeah, well at least they do no get in our way.

In order to fix the issue we need this part:

```
    # app label is deprecated and it will be removed after upgrading to Kubernetes 1.25
    app: "cluster-aws"
```

The rest would be to follow Helm guidelines and it is not causing any issues here.

It would be just a bit weird initially as we would have these two together:

```
    # app: "cluster-aws" label is deprecated and it will be removed after upgrading to Kubernetes 1.25
    app: "cluster-aws"
    app.kubernetes.io/name: "cluster"
```

But some deprecation message like that should clarify it.

Gacko:
Hm, I'm fine with that. Just take care of selector labels, they normally are immutable. But if I see correctly, there are no such labels in the chart.

nprokopic:
> Just take care of selector labels, they normally are immutable. But if I see correctly, there are no such labels in the chart.

Yeah, this is what prompted this PR initially.

Using cluster chart in cluster-aws is failing because the existing ClusterResourceSet already has a selector with `app: cluster-aws`, and in the cluster chart we are currently setting `app: cluster` and because of that ClusterResourceSet does not see the resources anymore with its selector, so we're fixing that here in the PR by changing `app: cluster` back to `app: cluster-aws`.

PR #42, opened 2024-01-09
Allow configuring cluster api server port

mnitchev:
### What does this PR do?

Adds a new value - `global.controlPlane.apiServer`. This configures the public facing port for the API on the LoadBalancer

New value, also changes the default from 6443 to 443 on CAPA.

### How does it look like?

N/A

https://github.com/giantswarm/roadmap/issues/3055

Make sure e2e tests pass. Agree on the value name.

It's a new value so probably yes?

Yes

### DIFF
--- a/helm/cluster/values.schema.json
+++ b/helm/cluster/values.schema.json
@@ -1291,7 +1291,8 @@
                             "type": "integer"
 =======
                             "type": "integer",
-                            "title": "The public facing API server Load Balancer port.",
+                            "title": "API server port",
+                            "description": "The public facing API server Load Balancer port.",
                             "default": 6443
                         },
                         "customNodeTaints": {
--- a/helm/cluster/values.schema.json
+++ b/helm/cluster/values.schema.json
@@ -139,8 +139,14 @@
                             "properties": {
                                 "storage": {
                                     "type": "object",
+<<<<<<< HEAD
                                     "title": "Storage",
                                     "description": "It describes the desired state of the system\u2019s storage devices.",
+||||||| parent of bbfacaf (Revert "run make generate-schema")
+=======
+                                    "title": "Storage",
+                                    "description": "It describes the desired state of the system’s storage devices.",
+>>>>>>> bbfacaf (Revert "run make generate-schema")
                                     "properties": {
                                         "directories": {
                                             "type": "array",
@@ -154,6 +160,7 @@
                                                     "path"
                                                 ],
                                                 "properties": {
+<<<<<<< HEAD
                                                     "filesystem": {
                                                         "type": "string",
                                                         "title": "Filesystem",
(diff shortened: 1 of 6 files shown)


mnitchev:
I don't know what's wrong with the schema. I ran `make generate-schema` and that just made things worse. If you tell me how to fix it I will.

Gacko:
`make generate-schema` uses a Helm plugin which auto-generates a schema JSON from the values YAML. The current version of the schema JSON looks manually maintained, so you could also just drop the whole `helm schema-gen` stuff from the Makefile.

Regarding the CI: Make `title` the description and add a title without a period: https://github.com/giantswarm/cluster/actions/runs/7541596783/job/20528649325?pr=42#step:3:150

helm/cluster/values.schema.json
```diff
@@ -744,6 +744,12 @@
                         "replicas"
                     ],
                     "properties": {
+                        "apiServerPort": {
+                            "type": "integer",
+                            "title": "API server port",
+                            "description": "The public facing API server Load Balancer port.",
```
Gacko, line 750: Maybe we should be more concrete on the load balancer thing. One could confuse this with the API server port.

mnitchev: I guess we can call it `apiServerLoadBalancerPort` then? The whole thing that's being set (i.e. `spec.ClusterNetwork.APIServerPort` in the Cluster CR) is very confusing to begin with and is used inconsistently, although I've only tested it on CAPA and just viewed the code for the other providers. Let me explain:
### CAPI
CAPI specifies this field as ["the port the API Server should bind to"](https://github.com/kubernetes-sigs/cluster-api/blob/1b662d33f8fe55511b47e0e8f9044c77acb848e4/api/v1beta1/cluster_types.go#L351-L354), however this isn't really used anywhere in the code as far as I can tell. The actual port that's being used comes from the KubeadmConfig. 

### CAPA
In CAPA (where I want to use this) the value is used to [configure the load balancer port **only**](https://github.com/kubernetes-sigs/cluster-api-provider-aws/blob/f8b9cfd43ff0dd3e2887c3acad3d1ac53284cf8c/pkg/cloud/services/elb/loadbalancer.go#L962-L969) (set [here](https://github.com/kubernetes-sigs/cluster-api-provider-aws/blob/f8b9cfd43ff0dd3e2887c3acad3d1ac53284cf8c/pkg/cloud/scope/cluster.go#L290-L295)). The destination port (i.e. the bindPort of the apiserver) is hardcoded to 6443. (Side note this is only for "classic" LBs, all others are hardcoded, but it's the default and is what we use anyway)

### CAPZ
CAPZ on the other hand sets [both the LB's port and the destination port ](https://github.com/kubernetes-sigs/cluster-api-provider-azure/blob/eadf744eb9ebbb7efa91bfed915104a23ed0abc8/azure/services/loadbalancers/spec.go#L229-L230C5). So I guess this means that you also need to set the `bindPort` when configuring the KubeadmControlPlane, but like I said I haven't tested this on anything other than CAPA.

### CAPV and CAPVCD

The value for `Spec.ClusterNetwork.APIServerPort` appears unused in those.

Gacko: Yeah, sounds confusing. I'd stick to how it is right now then.

nprokopic:
> I ran make generate-schema and that just made things worse

You sure you ran that? We don't have `make generate-schema` in cluster chart 😄

nprokopic:
`make generate-schema` is a leftover from when we started using the JSON schema across apps, as it was used to create the initial version of the JSON schema based on the current Helm values.yaml at that moment.

After initially creating the JSON schema, the changes are done in the schema and Helm values.yaml gets generated (CI also checks if this is done).

helm/cluster/values.schema.json
```diff
@@ -744,6 +744,12 @@
...
+                            "default": 6443
```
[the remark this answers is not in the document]
mnitchev: Done

nprokopic:
@mnitchev I added `make generate` in another merged PR, as a shortcut to run `make normalize-schema validate-schema generate-docs generate-values`.

If you rebase you can just run it and it should fix CI issues here.

mnitchev:
Man that was a pain in the ass to rebase. I really hate that the thing we have to deal with manually is in JSON while the thing we generate is in YAML. I wonder if it's possible to have the schema in a different format and then generate the JSON.

nprokopic:
> Man that was a pain in the ass to rebase. I really hate that the thing we have to deal with manually is in JSON while the thing we generate is in YAML.

Yeah, that one is a real PITA! One of the improvements that I would really like to do there soon is to break the schema into multiple reusable schemas (and put most of reusable schemas to https://github.com/giantswarm/schema).

> I wonder if it's possible to have the schema in a different format and then generate the JSON.

And about this, I was thinking to maybe try out https://cuelang.org/ in some way (if possbile), so we generate JSON schema from cuelang file. We see how it looks like and if people don't hate it and it makes our lives easier, we try to use it.

nprokopic (approved):

PR #43, opened 2024-01-10
Prepare control plane for AWS integration

nprokopic:
### What does this PR do?

This PR has changes that were recently added to cluster-aws and are now ported to cluster chart. It also has fixes that are needed to properly render control plane resources from cluster chart in cluster-aws.

Added:

- Enable using cgroupv1 (ported from https://github.com/giantswarm/cluster-aws/pull/410).
- Add systemd unit for OS hardening (ported from cluster-aws).
- Add systemd units for preventing in-place Flatcar OS updates (ported from cluster-aws).
- Add systemd unit for configuring kubeadm service (ported from cluster-aws).
- Add systemd unit for configuring containerd service (ported from cluster-aws).
- Add systemd unit for configuring audit-rules service (ported from cluster-aws).
- Add missing kubelet configuration to align it with vintage config (ported from https://github.com/giantswarm/cluster-aws/pull/468).

Changed:

- Support longer node pool names and allow dashes (ported from https://github.com/giantswarm/cluster-aws/pull/429).
- Use KubeletConfiguration file instead of a bash script (ported from https://github.com/giantswarm/cluster-aws/pull/427).
- Update kubernetes version to 1.24.16.
- Enable Teleport by default.
- Update Teleport version to 14.1.3.

cluster chart has the latest control plane changes from cluster-aws.

### How does it look like?

See diff.

https://github.com/giantswarm/roadmap/issues/3047


helm/cluster/files/etc/kubelet-configuration.yaml
```diff
@@ -0,0 +1,47 @@
...
+  memory: 1280Mi
+  ephemeral-storage: 1024Mi
+kubeReservedCgroup: /kubereserved.slice
+protectKernelDefaults: true
+systemReserved:
+  cpu: 250m
+  memory: 384Mi
+systemReservedCgroup: /system.slice
+tlsCipherSuites:
+- TLS_AES_128_GCM_SHA256
```
nprokopic, line 25: Compared to cluster-aws change, I have extended this list so we include all preferred TLS cipher suites that are listed in official [kubelet docs](https://v1-25.docs.kubernetes.io/docs/reference/command-line-tools-reference/kubelet/) (for CLI flag `--tls-cipher-suites`, but same values apply for `KubeletConfiguration`).

helm/cluster/files/etc/containerd/config.toml
```diff
@@ -19,7 +19,7 @@ no_shim = false
 # setting runc.options unsets parent settings
 runtime_type = "io.containerd.runc.v2"
 [plugins."io.containerd.grpc.v1.cri".containerd.runtimes.runc.options]
-SystemdCgroup = true
+SystemdCgroup = {{ if $.Values.internal.advancedConfiguration.cgroupsv1 }}false{{else}}true{{end}}
```
njuettner, line 22: Isn't it the other way around? I'm confused 😄

njuettner (approved):
Nice 👏🏻

helm/cluster/files/etc/containerd/config.toml
njuettner: Nevermind I misread it, sorry!

PR #47, opened 2024-01-17
Add script and service for k8s API server fairness

weseven:
Adds a systemd service and a script to be executed in controlplane machines to compute k8s API server fairness values based on machine resources (as happens in vintage).
Enable fairness in API server extraArgs. 

Towards: https://github.com/giantswarm/roadmap/issues/3003

### DIFF
--- a/helm/cluster/files/opt/bin/setup-apiserver-environment.sh
+++ b/helm/cluster/files/opt/bin/setup-apiserver-environment.sh
@@ -11,8 +11,8 @@ declare -i memory
 cpus=$((cpus*1000))
 
 # subtract dedicated millicores (see 'systemReserved' and 'kubeReserved' fields in the kubelet config file)
-sysreserved=$(echo {{ $.Values.internal.advancedConfiguration.kubelet.systemReserved.cpu | quote }} | tr -dc '0-9' )
-kubereserved=$( echo {{ $.Values.internal.advancedConfiguration.kubelet.kubeReserved.cpu | quote }} | tr -dc '0-9')
+sysreserved=$(echo {{ $.Values.internal.advancedConfiguration.kubelet.systemReserved.cpu | quote }} | tr -dc '0-9')
+kubereserved=$(echo {{ $.Values.internal.advancedConfiguration.kubelet.kubeReserved.cpu | quote }} | tr -dc '0-9')
 reserved=$((sysreserved + kubereserved))
 cpus=$((cpus-reserved))
 
--- a/helm/cluster/files/opt/bin/setup-apiserver-environment.sh
+++ b/helm/cluster/files/opt/bin/setup-apiserver-environment.sh
@@ -1,0 +1,28 @@
+#!/bin/bash
+env_file="/etc/apiserver-environment"
+
+cpus="$(grep -c ^processor /proc/cpuinfo)"
+memory="$(awk '/MemTotal/ { printf "%d \n", $2/1024/1024 }' /proc/meminfo)"
+
+declare -i cpus
+declare -i memory
+
+# convert to millicores
+cpus=$((cpus*1000))
+
+# subtract dedicated millicores (see 'systemReserved' and 'kubeReserved' fields in the kubelet config file)
+cpus=$((cpus-600))
+
+# we dedicate 3/4 of all millicores to the api server
+dedicated=$((cpus * 3 / 4))
+
+# for every core we have dedicated to api server we allow 200 requests for fairness.
+# 1/3 of such number is for mutating requests, 2/3 are for all requests.
+echo "MAX_REQUESTS_INFLIGHT=$((dedicated / 1000 * 200 * 2 / 3))" >$env_file
+echo "MAX_MUTATING_REQUESTS_INFLIGHT=$((dedicated / 1000 * 200 * 1 / 3 ))" >>$env_file
+echo "CPU_LIMIT=${dedicated}m" >>$env_file
+echo "MEMORY_LIMIT=$((memory * 3 / 4))Gi" >>$env_file
+
+# We request 1/3 of the available CPUs and 1/2 of the system memory just for api server.
+echo "CPU_REQUEST=$((cpus / 3))m" >>$env_file
+echo "MEMORY_REQUEST=$((memory / 2))Gi" >>$env_file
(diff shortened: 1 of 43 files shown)


helm/cluster/templates/clusterapi/controlplane/_helpers_flatcar.tpl
```diff
@@ -20,6 +20,18 @@ containerLinuxConfig:
 {{- if (((((($.Values.providerIntegration.controlPlane).kubeadmConfig).ignition).containerLinuxConfig).additionalConfig).systemd).units }}
 {{- include "cluster.internal.kubeadm.ignition.containerLinuxConfig.additionalConfig.systemd.units" $.Values.providerIntegration.controlPlane.kubeadmConfig.ignition.containerLinuxConfig.additionalConfig.systemd.units }}
 {{- end }}
+- name: setup-apiserver-environment.service
```
nprokopic, line 23: Let's put this unit to a new helper named `"cluster.internal.controlPlane.kubeadm.ignition.containerLinuxConfig.additionalConfig.systemd.units.default"` (in this same file) so it's done similarly to other cases.

Then it would be included above under:

```
include "cluster.internal.kubeadm.ignition.containerLinuxConfig.additionalConfig.systemd.units.default"
```

So we get (the one in the middle would be new):
```
containerLinuxConfig:
  additionalConfig: |
    systemd:
      units:
      {{- include "cluster.internal.kubeadm.ignition.containerLinuxConfig.additionalConfig.systemd.units.default" $ | indent 6 }}
      {{- include "cluster.internal.controlPlane.kubeadm.ignition.containerLinuxConfig.additionalConfig.systemd.units.default" $ | indent 6 }}
      {{- include "cluster.internal.controlPlane.kubeadm.ignition.containerLinuxConfig.additionalConfig.systemd.units" $ | indent 6 }}
```

This way we always render the systemd units in the following order:
1. default units for both control plane and worker nodes
2. default units for control plane nodes
3. provider-specific systemd units

So we get predictable and consistent resources across providers as well.

We have this similar ordering in many other places in the chart, the intent is still not documented, will try to fix that as much as possible.

nprokopic: P.S. Template names are too long I think, I'll open a PR some time next week to simplify them a bit.

e.g. change 

from

```
cluster.internal.controlPlane.kubeadm.ignition.containerLinuxConfig.additionalConfig.systemd.units
```

to

```
cluster.internal.controlPlane.ignition.systemd.units
```

or 

```
cluster.internal.controlPlane.os.systemd.units
```

or even shorter

```
cluster.internal.controlPlane.systemd.units
```

weseven: Makes a lot of sense, thanks!
I've moved it in the last commit :)

helm/cluster/files/opt/bin/setup-apiserver-environment.sh
```diff
@@ -0,0 +1,28 @@
...
+memory="$(awk '/MemTotal/ { printf "%d \n", $2/1024/1024 }' /proc/meminfo)"
+
+declare -i cpus
+declare -i memory
+
+# convert to millicores
+cpus=$((cpus*1000))
+
+# subtract dedicated millicores (see 'systemReserved' and 'kubeReserved' fields in the kubelet config file)
+cpus=$((cpus-600))
```
njuettner, line 14: Could we make this more flexible? We're using different CPU's, would there be a way to figure out how many CPU's we have on the node and do a calculation based on the amount of CPUs?

weseven: Not sure I understand your question, line 4 does exactly that: extracts the number of cpu (well, cores) available on the system.

njuettner: I mean the hardcoded 600

weseven: Clarifying: we might want to make the systemReserved and kubeReserved configurable.
At the moment the values are hardcoded here (that 600) and here in the kubelet config:
https://github.com/giantswarm/cluster/blob/0ab4c58d0b4ad004ef6b335afe38ee83feddb368/helm/cluster/files/etc/kubelet-configuration.yaml#L14-L21

@nprokopic , if we wanted to have those two as helm values, what would be the correct template?
Something like `.Values.providerIntegration.controlPlane.kubeadmConfig.ignition.containerLinuxConfig.additionalConfig.systemReserved.cpu` and `.Values.providerIntegration.controlPlane.kubeadmConfig.ignition.containerLinuxConfig.additionalConfig.kubeReserved.cpu` ?
And eventually, where's the correct template where to define the values and give defaults if not defined?

nprokopic: `.Values.providerIntegration` is supposed to be immutable provider-specific values, hard-coded in `values.schema.json` in cluster-<provider> apps.

I would put these for now here:

```
internal:
  advancedConfiguration:
    kubelet:
      kubeReserved: { ... }
      systemReserved: { ... }
```

Because those are customizable per cluster and have these disclaimers 😄:

```
For Giant Swarm internal use only, not stable, or not supported by UIs.

Advanced configuration of cluster components, to be configured by Giant Swarm staff only.
```

And it does seem that you should really know what you're doing if you are modifying kubelet's `kubeReserved` and/or `systemReserved` config.

Later if we want to make this available to customers we can move it to `.Values.global`.

nprokopic: I'm slowly adding docs in every new PR, so it should be more clear soon where to put what.

weseven: @nprokopic, added those values in the last commits and updated the schema and the generated docs and values (I fucked up and did a rebase instead of a ff merge from main, so github shows way more changes than it should, sorry :sweat:)

nprokopic: Sorry for delay, looking into it now.

nprokopic (approved):

weseven:
This is finally ready to be merged, tested in https://github.com/giantswarm/cluster-aws/pull/500

Last commits are an horror story in 4 characters, `'\""`, since I couldn't find a way to quote in the helm template in a way that preserves the quote when marshaling to json AND preserves the quotes when unmarshaling to the yml file AND preserves the quotes after substituting the environment variable with `envsubst`, WITHOUT adding single or double quotes since the end value should be an integer while also being a string and not an integer when marshalling to json.
I ended up quoting that way in the script so we get quotes directly in the env variable and we don't have to deal with the interaction between quote escaping in go templates, json, bash variables, and `'\""` seem to survive properly to `envsubst`. 
Pretty sure I was close to summoning one of the ancient ones in a couple of attempts.

PR #51, opened 2024-01-19
Move customisable advanced properties to Values.internal.advancedConfiguration

nprokopic:
### What does this PR do?

This PR moves advanced properties that can be customised for every cluster to `.Values.internal.advancedConfiguration`. These advanced properties are currently in `.Values.providerIntegration`, which is incorrect, because `.Values.providerIntegration` is meant to be immutable and set statically (hard-coded) only in Helm `values.yaml` in cluster-\<provider\> apps.

Changes:
- Move `extraCertificateSANs` property to `.Values.internal.advancedConfiguration.controlPlane.apiServer`.
- Move API server `extraArgs` Helm value to `.Values.internal.advancedConfiguration.controlPlane.apiServer`.
- Move API server `etcdPrefix` Helm value to `.Values.internal.advancedConfiguration.controlPlane.apiServer`.
- Move API server `bindPort` Helm value to `.Values.internal.advancedConfiguration.controlPlane.apiServer`.
- Add custom `files` config to Helm value to `.Values.internal.advancedConfiguration`.
- Add custom `preKubeadmCommands` config to Helm value to `.Values.internal.advancedConfiguration`.
- Add custom `postKubeadmCommands` config to Helm value to `.Values.internal.advancedConfiguration`.
- Move advanced etcd config to `.Values.internal.advancedConfiguration.controlPlane.etcd`.

Advanced cluster properties can be set in `.Values.internal.advancedConfiguration`

### How does it look like?

Before:

```yaml
providerIntegration:
  controlPlane:
    kubeadmConfig:
      clusterConfiguration:
        apiServer:
          etcdPrefix: giantswarm.io
          extraArgs:
            allow-privileged: false
            authorization-webhook-version: v1beta1
            delete-collection-workers: 2
            min-request-timeout: 600
          extraCertificateSANs:
          - this-is-extra-cert-san.giantswarm.io
          - another-extra-cert-san.giantswarm.io
        etcd: { ... }
      files:
      - path: /etc/hello/control-plane/node/stuff.yaml
        permissions: "0644"
        contentFrom:
          secret:
            name: cluster-super-secret-control-plane
            key: node-stuff
      localApiEndpoint:
        bindPort: 6443
```

After:

```yaml
internal:
  advancedConfiguration:
    controlPlane:
      apiServer:
        bindPort: 6443
        etcdPrefix: giantswarm.io
        extraArgs:
          allow-privileged: false
          authorization-webhook-version: v1beta1
          delete-collection-workers: 2
          min-request-timeout: 600
        extraCertificateSANs:
        - this-is-extra-cert-san.giantswarm.io
        - another-extra-cert-san.giantswarm.io
      etcd: { ... }
      files:
      - path: /etc/hello/control-plane/node/stuff.yaml
        permissions: "0644"
        contentFrom:
          secret:
            name: cluster-super-secret-control-plane
            key: node-stuff
```

We are customising these values via capi-migration-cli tool when migrating vintage AWS clusters to CAPA. Currently capi-migration-cli is setting these values in cluster-aws `Values.internal`, which will stop working when cluster-aws starts to render control plane resources from the cluster chart (https://github.com/giantswarm/cluster-aws/pull/479).

Auto-generated docs are updated.

### DIFF
--- a/helm/cluster/templates/clusterapi/controlplane/_helpers_clusterconfiguration_apiserver.tpl
+++ b/helm/cluster/templates/clusterapi/controlplane/_helpers_clusterconfiguration_apiserver.tpl
@@ -49,7 +49,7 @@ extraArgs:
   service-account-lookup: "true"
   service-cluster-ip-range: {{ .Values.global.connectivity.network.services.cidrBlocks | first }}
   tls-cipher-suites: {{ include "cluster.internal.controlPlane.kubeadm.clusterConfiguration.apiServer.tlsCipherSuites" $ }}
-  {{- range $argName, $argValue := ((($.Values.providerIntegration.controlPlane.kubeadmConfig).clusterConfiguration).apiServer).extraArgs }}
+  {{- range $argName, $argValue := $.Values.internal.advancedConfiguration.controlPlane.apiServer.extraArgs }}
   {{ $argName }}: {{ if kindIs "string" $argValue }}{{ $argValue | quote }}{{ else }}{{ $argValue }}{{ end }}
   {{- end }}
 extraVolumes:
--- a/helm/cluster/templates/clusterapi/controlplane/_helpers_clusterconfiguration_apiserver.tpl
+++ b/helm/cluster/templates/clusterapi/controlplane/_helpers_clusterconfiguration_apiserver.tpl
@@ -25,8 +25,8 @@ extraArgs:
   cloud-provider: external
   enable-admission-plugins: {{ include "cluster.internal.controlPlane.kubeadm.clusterConfiguration.apiServer.enableAdmissionPlugins" $ }}
   encryption-provider-config: /etc/kubernetes/encryption/config.yaml
-  {{- if $.Values.providerIntegration.controlPlane.kubeadmConfig.clusterConfiguration.apiServer.etcdPrefix }}
-  etcd-prefix: {{ $.Values.providerIntegration.controlPlane.kubeadmConfig.clusterConfiguration.apiServer.etcdPrefix }}
+  {{- if $.Values.internal.advancedConfiguration.controlPlane.apiServer.etcdPrefix }}
+  etcd-prefix: {{ $.Values.internal.advancedConfiguration.controlPlane.apiServer.etcdPrefix }}
   {{- end }}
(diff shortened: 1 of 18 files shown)


helm/cluster/templates/clusterapi/controlplane/_helpers_clusterconfiguration_apiserver.tpl
```diff
@@ -25,8 +25,8 @@ extraArgs:
   cloud-provider: external
   enable-admission-plugins: {{ include "cluster.internal.controlPlane.kubeadm.clusterConfiguration.apiServer.enableAdmissionPlugins" $ }}
   encryption-provider-config: /etc/kubernetes/encryption/config.yaml
-  {{- if $.Values.providerIntegration.controlPlane.kubeadmConfig.clusterConfiguration.apiServer.etcdPrefix }}
-  etcd-prefix: {{ $.Values.providerIntegration.controlPlane.kubeadmConfig.clusterConfiguration.apiServer.etcdPrefix }}
+  {{- if $.Values.internal.advancedConfiguration.controlPlane.apiServer.etcdPrefix }}
+  etcd-prefix: {{ $.Values.internal.advancedConfiguration.controlPlane.apiServer.etcdPrefix }}
```
njuettner, line 29: If we anyway changing the contract a bit, WDYT about moving the etcdPrefix into etcd?

It's a flag on the API server but at the same time it's ETCD configuration. 
Maybe it's better to have all ETCD configurations under one roof (etcd)

nprokopic: IIUC it's actual API server config, and how API server uses etcd as a client/consumer (`--etcd-prefix` flag is passed to API server, not to etcd, you can find it [here](https://kubernetes.io/docs/reference/command-line-tools-reference/kube-apiserver/) in API server CLI flags docs), i.e. it's about how the API server will name the resources it creates for itself.

Some other other etcd client would use another prefix, so different clusters can use different prefixes and then share the same etcd cluster (e.g. see this example issue [here](https://github.com/kubernetes/kubeadm/issues/1503)).

The name of flag `--etcd-prefix` is not the best TBH 😄 it's a bit misleading.

PR #50, opened 2024-01-19
Improve NO_PROXY template

nprokopic:
Towards https://github.com/giantswarm/roadmap/issues/3047

### What does this PR do?

Coming from https://github.com/giantswarm/cluster-aws/pull/479#discussion_r1455411843

This PR does few things:
- It renames NO_PROXY template from `cluster.internal.kubeadm.proxy.noProxyList` to `cluster.connectivity.proxy.noProxy`
  - So this named template is not internal anymore, and now it can be considered public API offered by the cluster chart.
- It refactors the `cluster.connectivity.proxy.noProxy` template so it can be correctly called from other charts (e.g. render NO_PROXY value in cluster-aws).

### What is the effect of this change to users?

Developers from provider-integration KaaS teams can use `cluster.connectivity.proxy.noProxy` template to render NO_PROXY value consistently in their charts.

### How does it look like?

Example usage when inlcuded in the cluster chart:

```
{{ include "cluster.connectivity.proxy.noProxy" (dict "global" $.Values.global "providerIntegration" $.Values.providerIntegration) }}
```

Example usage when inlcuded in the parent chart (cluster-\<provider\> app):

```
{{ include "cluster.connectivity.proxy.noProxy" (dict "global" $.Values.global "providerIntegration" $.Values.cluster.providerIntegration) }}
```

### Any background context you can provide?

PR review in cluster-aws where Nick spotted an issue with NO_PROXY value rendering https://github.com/giantswarm/cluster-aws/pull/479#discussion_r1455411843

### Do the docs need to be updated?

Yes. Updated Helm template docs.

### Should this change be mentioned in the release notes?

- [x] CHANGELOG.md has been updated (if it exists)

PR #53, opened 2024-01-22
Fix cluster.connectivity.proxy.noProxy template

nprokopic:
### What does this PR do?

Fix "cluster.connectivity.proxy.noProxy" template to correctly render values from specified template (a bug was introduced in https://github.com/giantswarm/cluster/pull/50).

NO_PROXY is correctly rendered.

https://github.com/giantswarm/cluster-aws/pull/479#discussion_r1455411843

### DIFF
--- a/helm/cluster/templates/clusterapi/_helpers.tpl
+++ b/helm/cluster/templates/clusterapi/_helpers.tpl
@@ -51,7 +51,8 @@
 {{- end }}
 {{- /* Add provider-specific NO_PROXY values from template */}}
 {{- if $providerIntegration.connectivity.proxy.noProxy.templateName }}
-{{- range $noProxyAddress := include $providerIntegration.connectivity.proxy.noProxy.templateName $ | fromYamlArray }}
+{{- $values := dict "global" $global }}
+{{- range $noProxyAddress := include $providerIntegration.connectivity.proxy.noProxy.templateName (dict "Values" $values) | fromYamlArray }}
 {{- $noProxyList = append $noProxyList $noProxyAddress -}}
 {{- end }}
 {{- end }}
@@ -67,7 +68,7 @@
 {{- end }}
 
 {{- define "cluster.test.internal.kubeadm.proxy.anotherNoProxyList" }}
-- some.noproxy.address.giantswarm.io
+- some.noproxy.{{ $.Values.global.metadata.name }}.{{ $.Values.global.connectivity.baseDomain }}
 - another.noproxy.address.giantswarm.io
 {{- end }}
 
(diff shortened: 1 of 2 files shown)


helm/cluster/templates/clusterapi/_helpers.tpl
```diff
@@ -51,7 +51,8 @@
 {{- end }}
 {{- /* Add provider-specific NO_PROXY values from template */}}
 {{- if $providerIntegration.connectivity.proxy.noProxy.templateName }}
-{{- range $noProxyAddress := include $providerIntegration.connectivity.proxy.noProxy.templateName $ | fromYamlArray }}
+{{- $values := dict "global" $global }}
+{{- range $noProxyAddress := include $providerIntegration.connectivity.proxy.noProxy.templateName (dict "Values" $values) | fromYamlArray }}
```
nprokopic, line 55: Template that we include needs global Helm values.

nprokopic:
There is a rendering diff because I updated the test (dummy) template.

helm/cluster/templates/clusterapi/_helpers.tpl
njuettner, line 55: Just for my understanding, the name "Values" in `dict "Values" $values` is arbitrary right? 

It could be also named "ProxyList" or anything else right?

njuettner (approved):

helm/cluster/templates/clusterapi/_helpers.tpl
nprokopic: Actually no, it's to enable called template to use `.Values`.

e.g. here, template `"cluster.connectivity.proxy.noProxy"` calls another  template by name like this:

```
include $providerIntegration.connectivity.proxy.noProxy.templateName (dict "Values" $values)
```

e.g. for $providerIntegration.connectivity.proxy.noProxy.templateName value cluster-aws will pass `"awsNoProxyList"` template name, so we call template like that with that `dict` because `"awsNoProxyList"` template looks like this (it uses `$.Values`):

```
{{- define "awsNoProxyList" }}
- {{ $.Values.global.connectivity.network.vpcCidr }}
{{- end }}
```

So the caller of `"awsNoProxyList"` has to pass an argument (object/map) that has `.Values`

nprokopic: Initially the call was made plainly like this:

```
include $providerIntegration.connectivity.proxy.noProxy.templateName $
```

so there was a Helm templating error like this:

```
executing "cluster.connectivity.proxy.noProxy" at <include $providerIntegration.connectivity.proxy.noProxy.templateName $>:

error calling include: template: cluster-aws/templates/_helpers.tpl:372:6:

executing "awsNoProxyList" at <$.Values.global.connectivity.network.vpcCidr>:

nil pointer evaluating interface {}.global
```

PR #55, opened 2024-01-24
Add etcd quota backend bytes config

njuettner:
Towards: https://github.com/giantswarm/roadmap/issues/2977

We want to add the possibility to adjust the etcd quota backend bytes up to 16GiB.

### DIFF
--- a/helm/cluster/values.schema.json
+++ b/helm/cluster/values.schema.json
@@ -1152,11 +1152,11 @@
                                                 "existing"
                                             ]
                                         },
-                                        "quotaBackendBytes": {
-                                            "type": "string",
-                                            "title": "Quota backend bytes",
+                                        "quotaBackendBytesGiB": {
+                                            "type": "integer",
+                                            "title": "Quota backend bytes in GiB",
                                             "description": "Raise the etcd default backend bytes limit up to 16GiB.",
-                                            "default": "8589934592"
+                                            "default": 8
                                         }
                                     }
                                 },
--- a/helm/cluster/values.schema.json
+++ b/helm/cluster/values.schema.json
@@ -1156,7 +1156,9 @@
                                             "type": "integer",
                                             "title": "Quota backend bytes in GiB",
                                             "description": "Raise the etcd default backend bytes limit up to 16GiB.",
-                                            "default": 8
+                                            "default": 8,
+                                            "maximum": 16,
+                                            "minimum": 8
                                         }
                                     }
                                 },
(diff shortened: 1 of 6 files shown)


helm/cluster/values.schema.json
```diff
@@ -1151,6 +1151,12 @@
                                                 "new",
                                                 "existing"
                                             ]
+                                        },
+                                        "quotaBackendBytes": {
+                                            "type": "string",
+                                            "title": "Quota backend bytes",
+                                            "description": "Raise the etcd default backend bytes limit up to 16GiB.",
+                                            "default": "8589934592"
```
njuettner, line 1159: Unfortunately I need to use a string here, using integer will add a nasty `quota-backend-bytes: 8.589934592e+09` to values

nprokopic: Do you know if it possible to set values in gigabytes, e.g. `16g`, instead of setting bytes? We can have a flag named `quotaBackendBytesGiB`, with some JSON schema validation like `minimum: 8` and `maximum: 16` (I can dig out the Slack thread about recommended min and max), so it's easier to set values without potential errors.

If etcd does not directly accept values in GiB like `16g`, then in the template we can just multiply config Helm value when setting the etcd flag value in the resource, so something like:

```
--quota-backend-bytes: {{ mul $etcdConfig.quotaBackendBytesGiB 1024 1024 1024 }}
```

So we have a simple to use API here, where we set a easy-to-type-and-think-about number in 8-16 range, instead of a number with 11 digits :see_no_evil:

njuettner: No it's not possible etcd only accepts int(64).

I already picked the number of min and max from the thread 8GiB and max 16GiB.

I think using `{{ mul $etcdConfig.quotaBackendBytesGiB 1024 1024 1024 }}` is super useful, I'll change it

nprokopic: > We can have a flag named quotaBackendBytesGiB, with some JSON schema validation like minimum: 8 and maximum: 16

This way to value would be `int`, so we could have `minimum` and `maximum` validations in the schema.

And we could just set the value directly without doing this in the template:
```
{{- if gt (int $etcdConfig.quotaBackendBytes) 17179869184 }}
```

Here above, if somebody sets something even higher in `$etcdConfig.quotaBackendBytes`, we silently fallback to a hard-coded max value here (17179869184).

With JSON schema validations you would immediately get a Helm error when you try to apply the chart with invalid value that does not respect `minimum` or `maximum`.

nprokopic (approved):

PR #60, opened 2024-01-25
Added support for RolloutBefore field in KubeadmControlPlane

vvondruska:
### What does this PR do?

This PR introduces a new field in the Helm values called `rolloutBefore`. It can be used to control how many days before the expiration of control plane certificates should the control plane nodes be rolled. This mechanism is used by CAPI to ensure that control plane certificates are renewed automatically.

It ensures that certificates on control plane nodes are rotated automatically.

### How does it look like?

There is a new field available in Values:

```yaml
internal:
  advancedConfiguration:
    controlPlane:
      rolloutBefore: # this structure is taken from CAPI
        certificatesExpiryDays: 180 # 180 is the default value
```

It is added to the `KubeadmControlPlane` CR.

Towards: https://github.com/giantswarm/roadmap/issues/3147

Check if the new property located in the correct section of the Helm Values and possibly also test that it works.

Yes

Probably yes

### DIFF
--- a/helm/cluster/templates/clusterapi/controlplane/kubeadmcontrolplane.yaml
+++ b/helm/cluster/templates/clusterapi/controlplane/kubeadmcontrolplane.yaml
@@ -58,5 +58,7 @@ spec:
     {{- $users | indent 4 }}
     {{- end }}
   replicas: {{ .Values.global.controlPlane.replicas }}
+  rolloutBefore:
+    {{- toYaml $.Values.global.controlPlane.rolloutBefore | nindent 4 }}
   version: v{{ trimPrefix "v" .Values.providerIntegration.kubernetesVersion }}
 {{- end }}
--- a/helm/cluster/templates/clusterapi/controlplane/kubeadmcontrolplane.yaml
+++ b/helm/cluster/templates/clusterapi/controlplane/kubeadmcontrolplane.yaml
@@ -57,8 +57,8 @@ spec:
     users:
     {{- $users | indent 4 }}
     {{- end }}
-  replicas: {{ .Values.global.controlPlane.replicas }}
+  replicas: {{ .Values.internal.advancedConfiguration.controlPlane.replicas }}
   rolloutBefore:
-    {{- toYaml $.Values.global.controlPlane.rolloutBefore | nindent 4 }}
+    {{- toYaml $.Values.internal.advancedConfiguration.controlPlane.rolloutBefore | nindent 4 }}
   version: v{{ trimPrefix "v" .Values.providerIntegration.kubernetesVersion }}
 {{- end }}
--- a/helm/cluster/templates/clusterapi/controlplane/kubeadmcontrolplane.yaml
+++ b/helm/cluster/templates/clusterapi/controlplane/kubeadmcontrolplane.yaml
@@ -57,7 +57,7 @@ spec:
     users:
     {{- $users | indent 4 }}
     {{- end }}
-  replicas: {{ .Values.internal.advancedConfiguration.controlPlane.replicas }}
+  replicas: {{ .Values.global.controlPlane.replicas }}
   rolloutBefore:
     {{- toYaml $.Values.internal.advancedConfiguration.controlPlane.rolloutBefore | nindent 4 }}
   version: v{{ trimPrefix "v" .Values.providerIntegration.kubernetesVersion }}
--- a/helm/cluster/values.schema.json
+++ b/helm/cluster/values.schema.json
@@ -894,7 +894,9 @@
                             "type": "object",
                             "title": "Rollout Before",
                             "description": "Rollout Before is a field to indicate a rollout should be performed if the specified criteria is met",
-                            "required": ["certificatesExpiryDays"],
+                            "required": [
+                                "certificatesExpiryDays"
+                            ],
                             "properties": {
                                 "certificatesExpiryDays": {
                                     "type": "integer",
--- a/helm/cluster/values.schema.json
+++ b/helm/cluster/values.schema.json
@@ -893,7 +893,7 @@
                         "rolloutBefore": {
                             "type": "object",
                             "title": "Rollout Before",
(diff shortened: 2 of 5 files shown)


helm/cluster/templates/clusterapi/controlplane/kubeadmcontrolplane.yaml
```diff
@@ -58,5 +58,7 @@ spec:
     {{- $users | indent 4 }}
     {{- end }}
   replicas: {{ .Values.global.controlPlane.replicas }}
+  rolloutBefore:
```
njuettner, line 61: The usage of `Values.global` usually indicates customer is allowed to change it, is that something we really need or is that something only GS should do?

njuettner:
You might wanna also run `make generate-docs`

helm/cluster/templates/clusterapi/controlplane/kubeadmcontrolplane.yaml
vvondruska: Good question. This property controls how long before the control plane certificates expiration should the nodes be rolled (and certificates renewed).

I understand that some customers need to configure their own time periods to renew the certificates (and roll the nodes in this case).

In vintage the detault time before the certificate expiration when the rotation should be triggered is set to 180 days. But some customers need to rotate their certificates e.g. every 3 months.

vvondruska: Do you know if the customers are able to control this in the vintage clusters?

njuettner: Ok thanks, the reason why I'm asking is there's a difference between `global` and `internal` 

`providerIntegration` is a static one and should never be changed
`internal` value represents only GS employees should change it
`global` is aiming customer values

vvondruska: Apparently customers did not need to modify this configuration directly in the past, so we can make it internal.

vvondruska: Should it go to `internal.advancedConfiguration.controlPlane`?

nprokopic: > Apparently customers did not need to modify this configuration directly in the past

If customers do not need to modify this flag, then we don't even need a Helm value for it and it can be static. We can always make it configurable later and expose a Helm value.

nprokopic: If we really (not maybe) know about some advanced use case where somebody from Giant Swarm has to tweak this config for some customer, then we can put it to `internal.advancedConfiguration.controlPlane.*`

nprokopic: Basically, we want to avoid exposing Helm values if we don't have to do it. The smaller API (exposed Helm) the better.

njuettner: Thinking about it, I think it would be best to let customers decide.

You made a good point, it might better make this flexible. In case customer need to follow a policy which requires rolling certificates every 90 days or so it might be best to giving them the possibility

vvondruska: I think we need to configure it in some cases (e.g. in case the customer needs a shorter renewal period), but it's OK if we set this property internally as GS.

I discussed it with Tobias from Team Phoenix and in his opinion it's better to make the property internal.

vvondruska: I moved the `rolloutBefore` property to `internal.advancedConfiguration.controlPlane` and added validation,

helm/cluster/values.schema.json
```diff
@@ -1186,6 +1189,24 @@
...
+                                        "certificatesExpiryDays"
+                                    ],
+                                    "properties": {
+                                        "certificatesExpiryDays": {
+                                            "type": "integer",
+                                            "title": "Certificate expiry days",
+                                            "description": "Indicates a rollout needs to be performed if the certificates of the machine will expire within the specified days.",
+                                            "default": 180,
+                                            "maximum": 364,
+                                            "minimum": 30
```
vvondruska, line 1207: CAPI docs recommends to set a long enough period of time before the certificates expire, so that there is enough time for the rotation to complete. 1 month should be enough time.

helm/cluster/values.schema.json
```diff
@@ -1186,6 +1189,24 @@
...
+                                    "required": [
+                                        "certificatesExpiryDays"
+                                    ],
+                                    "properties": {
+                                        "certificatesExpiryDays": {
+                                            "type": "integer",
+                                            "title": "Certificate expiry days",
+                                            "description": "Indicates a rollout needs to be performed if the certificates of the machine will expire within the specified days.",
+                                            "default": 180,
+                                            "maximum": 364,
```
vvondruska, line 1206: Certificates in CAPI clusters expire after a year by default. If the `certificatesExpiryDays` value is set to 365 days before the certificate expires, then the nodes are rolled continuously and the rollout never stops. The maximum value of 364 days ensures that certificates are not rotated for at least a day.

In theory it is possible to set an expiration period shorter or longer than 1 year. In that case the maximum value can be restrictive. If the certificate expiration period is set to 2 years, then the maximum value will make it impossible to trigger the rollout more than 1 year before the certificates expire. 

Question is if there is a need for such use case - certificates valid for 2 years rotated more than a year before they expire.

vvondruska:
E2E test results: https://github.com/giantswarm/cluster-aws/pull/503

helm/cluster/values.schema.json
nprokopic: IMO this here is more than good enough for now.

nprokopic (approved):

PR #68, opened 2024-01-31
Fix MachineHealthCheck annotation rendering

nprokopic:

### What does this PR do?

Fix MachineHealthCheck annotation rendering issue discovered in https://github.com/giantswarm/cluster-aws/pull/497

MachineHealthCheck annotation rendering works correctly when custom annotations are not specified.

### How does it look like?

See diff.

https://github.com/giantswarm/cluster-aws/pull/497

### DIFF
--- a/Makefile.development.mk
+++ b/Makefile.development.mk
@@ -7,10 +7,16 @@ ensure-schema-gen:
 
 ##@ Build
 
+TEST_CASE ?=
+ifdef TEST_CASE
+CI_FILE = "ci/test-$(TEST_CASE)-values.yaml"
+else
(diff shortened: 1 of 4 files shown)


Makefile.development.mk
```diff
@@ -7,10 +7,16 @@ ensure-schema-gen:
 
 ##@ Build
 
+TEST_CASE ?=
+ifdef TEST_CASE
+CI_FILE = "ci/test-$(TEST_CASE)-values.yaml"
```
nprokopic, line 12: small addition so we can locally run something like this:

```bash
TEST_CASE=required make template
```

weseven (approved):

PR #70, opened 2024-02-02
Fix node pool templates

nprokopic:
Towards https://github.com/giantswarm/roadmap/issues/3194

### What does this PR do?

This PR ports cluster-aws node pool fix from https://github.com/giantswarm/cluster-aws/pull/457.

Other fixes:
- Include `/var/lib/kubelet` directly by default in Ignition on worker nodes.
- Include `/etc/apiserver-environment` environment file in `kubeadm` systemd unit only on control plane nodes.

Node pools will roll correctly when an upgrade that requires that is performed.

### How does it look like?

See diff.

- https://github.com/giantswarm/cluster-aws/pull/457
- https://github.com/giantswarm/roadmap/issues/3194

### DIFF
--- a/helm/cluster/templates/clusterapi/workers/machinepool.yaml
+++ b/helm/cluster/templates/clusterapi/workers/machinepool.yaml
@@ -1,6 +1,7 @@
 {{- if $.Values.providerIntegration.resourcesApi.machinePoolResourcesEnabled }}
 {{- if eq $.Values.providerIntegration.resourcesApi.nodePoolKind "MachinePool" }}
 {{- range $nodePoolName, $nodePoolConfig := $.Values.global.nodePools | default $.Values.providerIntegration.workers.defaultNodePools }}
+{{- $_ := set $ "nodePool" (dict "name" $nodePoolName "config" $nodePoolConfig) }}
 apiVersion: cluster.x-k8s.io/v1beta1
 kind: MachinePool
 metadata:
--- a/helm/cluster/templates/clusterapi/workers/machinepool.yaml
+++ b/helm/cluster/templates/clusterapi/workers/machinepool.yaml
@@ -1,11 +1,10 @@
 {{- if $.Values.providerIntegration.resourcesApi.machinePoolResourcesEnabled }}
 {{- if eq $.Values.providerIntegration.resourcesApi.nodePoolKind "MachinePool" }}
-{{- range $nodePoolName, $nodePoolConfig := $.Values.global.nodePools }}
+{{- range $nodePoolName, $nodePoolConfig := $.Values.global.nodePools | default $.Values.providerIntegration.workers.defaultNodePools }}
 apiVersion: cluster.x-k8s.io/v1beta1
 kind: MachinePool
 metadata:
   annotations:
-    helm.sh/resource-policy: keep
     machine-pool.giantswarm.io/name: {{ include "cluster.resource.name" $ }}-{{ $nodePoolName }}
     cluster.x-k8s.io/replicas-managed-by: "external-autoscaler"
     {{- include "cluster.annotations.custom" $ | indent 4 }}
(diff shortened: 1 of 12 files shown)


helm/cluster/templates/clusterapi/workers/machinepool.yaml
```diff
@@ -1,11 +1,11 @@
 {{- if $.Values.providerIntegration.resourcesApi.machinePoolResourcesEnabled }}
 {{- if eq $.Values.providerIntegration.resourcesApi.nodePoolKind "MachinePool" }}
-{{- range $nodePoolName, $nodePoolConfig := $.Values.global.nodePools }}
+{{- range $nodePoolName, $nodePoolConfig := $.Values.global.nodePools | default $.Values.providerIntegration.workers.defaultNodePools }}
```
njuettner, line 3: This is a bit confusing to me, but why do we sometimes use nodePool and sometimes machinePool as variables.

Could we try to align here if that is possible?

helm/cluster/templates/clusterapi/workers/machinepool.yaml
```diff
@@ -27,14 +27,14 @@ metadata:
...
-  replicas: {{ $nodePoolConfig.replicas }}
+  replicas: {{ $nodePoolConfig.minSize | default $nodePoolConfig.replicas }}
   template:
     spec:
       bootstrap:
         configRef:
           apiVersion: bootstrap.cluster.x-k8s.io/v1beta1
           kind: KubeadmConfig
-          name: {{ include "cluster.resource.name" $ }}-{{ $nodePoolName }}
+          name: {{ include "cluster.resource.name" $ }}-{{ $nodePoolName }}-{{ include "cluster.internal.workers.kubeadmConfig.spec.hash" $ }}
```
njuettner, line 37: Why do we need to add an hash here?

helm/cluster/templates/clusterapi/workers/machinepool.yaml
nprokopic: Yeah, good point, will check and align.

helm/cluster/templates/clusterapi/workers/machinepool.yaml
nprokopic: TL;DR would be that there is bug where AWSMachinePool nodes do not do rolling update when referenced `KubeadmConfig` changes, so here we append the `KubeadmConfig` hash to the CR name, so when KubeadmConfig spec changes the CR name changes as well, and AWSMachinePool gets a new KubeadmConfig, which correctly triggers a rolling update. (we already do something similar for the control plane as well)

I have ported [this change](https://github.com/giantswarm/cluster-aws/pull/457) from cluster-aws, the explanation (copied from cluster-aws as well) with issue links is here in the PR https://github.com/giantswarm/cluster/pull/70/files#diff-bb8c36a7911fe46567a52f4b7ba28fe0bba14a08572f2250180eb1917a800454R2

njuettner: Oh wow, thank you.

njuettner (approved):
Not blocking here

helm/cluster/templates/clusterapi/workers/machinepool.yaml
nprokopic: I renamed `$machinePools` to `$nodePools` in one of the helper files.

So now we have:
- `$nodePools` as a variable for our Helm values, as we are using the term node pools in our API.
- We have MachinePools as a specific node pools implementation, so the term MachinePool is used in places where it refers to CAPI's MachinePool CR.
- We will at some point (soon) add MachineDeployments as another node pools implementation, and the naming should work like above - `$nodePools` as a variable name and MachineDeployment when referring to CAPI MachineDeployments specifically.

njuettner: Wonderful ❤️ ❤️ ❤️

PR #86, opened 2024-02-15
Support prepending cluster name to file secret name

AndiDog:
### What does this PR do?

Towards https://github.com/giantswarm/roadmap/issues/3183

Tested together with my cluster-aws change to introduce ENI mode for Cilium.

That requires configuring a certain dynamic config file on the host. We found that using the `contentFrom` (= from Secret) feature was the only reasonable option to pass the file content to the `cluster` chart.

None, since the new field is optional. A value of `contentFrom.secret.prependClusterNameAsPrefix=true` transforms the secret name. This is because the parent chart (e.g. `cluster-aws`) cannot pass dynamic values down that easily, so prepending the already-known cluster name is an easy solution. We have many objects whose name is `<cluster name>-something`, so this is quite usual.

### DIFF
--- a/helm/cluster/templates/clusterapi/_helpers_files.tpl
+++ b/helm/cluster/templates/clusterapi/_helpers_files.tpl
@@ -139,7 +139,7 @@ and is used to join the node to the teleport cluster.
 {{- $clusterName := required "clusterName is required for cluster.processFiles function call" .clusterName }}
 {{- $outFiles := list }}
 {{- range $file := .files }}
-{{- if eq (default false (index $file "contentFrom" "secret" "prependClusterNameAsPrefix")) true }}
+{{- if default false (index $file "contentFrom" "secret" "prependClusterNameAsPrefix") }}
 {{- $secret := (index $file "contentFrom" "secret") }}
 {{- $secretName := (required "Secret name must be given" (index $secret "name")) }}
 {{- $_ := set $secret "name" (printf "%s-%s" $clusterName $secretName) }}
--- a/helm/cluster/templates/clusterapi/_helpers_files.tpl
+++ b/helm/cluster/templates/clusterapi/_helpers_files.tpl
@@ -124,13 +124,29 @@ and is used to join the node to the teleport cluster.
 {{/* Provider-specific files for all nodes */}}
 {{- define "cluster.internal.kubeadm.files.provider" }}
 {{- if $.Values.providerIntegration.kubeadmConfig.files }}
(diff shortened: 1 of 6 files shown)


helm/cluster/templates/clusterapi/_helpers_files.tpl
```diff
@@ -124,13 +124,29 @@ and is used to join the node to the teleport cluster.
...
-{{ toYaml $.Values.internal.advancedConfiguration.files }}
+{{ include "cluster.processFiles" (dict "files" $.Values.internal.advancedConfiguration.files "clusterName" (include "cluster.resource.name" $)) }}
 {{- end }}
 {{- end }}
+
+{{- define "cluster.processFiles" }}
+{{- $clusterName := required "clusterName is required for cluster.processFiles function call" .clusterName }}
+{{- $outFiles := list }}
+{{- range $file := .files }}
+{{- if eq (default false (index $file "contentFrom" "secret" "prependClusterNameAsPrefix")) true }}
```
Gacko, line 142:
```suggestion
{{- if $file.contentFrom.secret.prependClusterNameAsPrefix | default false }}
```

Gacko: Or something like that. At least you don't need an `eq` here.

AndiDog: Right, fixed that. But the `index` call is needed because that supports chaining on null values instead of failing. And I'm keeping the `default false (index ...)` to denote that this expression is checking for a boolean (and isn't mean to do `hasKey` instead).

helm/cluster/templates/clusterapi/_helpers_files.tpl
```diff
@@ -124,13 +124,29 @@ and is used to join the node to the teleport cluster.
...
 
 {{/* Custom cluster-specific files for all nodes */}}
 {{- define "cluster.internal.kubeadm.files.custom" }}
 {{- if $.Values.internal.advancedConfiguration.files }}
-{{ toYaml $.Values.internal.advancedConfiguration.files }}
+{{ include "cluster.processFiles" (dict "files" $.Values.internal.advancedConfiguration.files "clusterName" (include "cluster.resource.name" $)) }}
 {{- end }}
 {{- end }}
+
+{{- define "cluster.processFiles" }}
```
njuettner, line 138: Just an idea but maybe it would make sense to align with the other define functions here and name it `cluster.internal.processFiles`? Thoughts @nprokopic ?

njuettner (approved):
not blocking here, seems ok to me

PR #90, opened 2024-02-19
Move apiserver environment generation from systemd service to prekubeadm commands & kubeadm patches

weseven:
### What does this PR do?

Removes the setup-apiserver-environment systemd service.
Instead it uses preKubeadmCommands on the controlplane to execute the script generating the environment variables and then substituting the placeholders variables in a patch file in /etc/kubernetes/patches.
The patch file adds the two args `--max-requests-inflight` and `--max-mutating-requests-inflight` to the apiserver args.
Patches are applied on init/join from nodes in kubeadm, and if we change node size the proper values are computed and applied.

Hopefully this solves some weird race conditions that in some cases prevented the proper variable substitutions, e.g., https://github.com/giantswarm/cluster-aws/pull/504, https://gigantic.slack.com/archives/C0559SH3RJ4/p1707994275400329


AndiDog (approved):
`xargs` could be brittle, but as implementation detail it's fine if you think it's good enough

helm/cluster/templates/clusterapi/controlplane/_helpers_prekubeadmcommands.tpl
```diff
@@ -9,10 +9,18 @@
...
+{{- include "cluster.internal.controlPlane.kubeadm.preKubeadmCommands.default" $ }}
 {{- include "cluster.internal.controlPlane.kubeadm.preKubeadmCommands.provider" $ }}
 {{- include "cluster.internal.controlPlane.kubeadm.preKubeadmCommands.custom" $ }}
 {{- end }}
 
+{{/* Default commands to run before kubeadm on control plane nodes */}}
+{{- define "cluster.internal.controlPlane.kubeadm.preKubeadmCommands.default" }}
+- /opt/bin/setup-apiserver-environment.sh
+- env $(cat /etc/apiserver-environment | xargs) envsubst < /etc/kubeadm.yml > /etc/kubeadm.yml.tmp
+- mv /etc/kubeadm.yml.tmp /etc/kubeadm.yml
```
njuettner, line 21: do we need to change the file permission or overwriting owner / group again or is this ok?

njuettner (approved):

helm/cluster/templates/clusterapi/controlplane/_helpers_prekubeadmcommands.tpl
weseven: It's ok, the original file as well is owned by root (these are the same instructions we use in [the flatcarKubeadmPreCommands in cluster-aws](https://github.com/giantswarm/cluster-aws/blob/13e88385966bf2ac69b076a819b84f4a8975804a/helm/cluster-aws/templates/_helpers.tpl#L332)

weseven:
> `xargs` could be brittle, but as implementation detail it's fine if you think it's good enough

True, potential footgun, but we are the ones writing the output file, it should be ok.

PR #92, opened 2024-02-20
Kubelet: Add `containerLogMaxSize` & `containerLogMaxFiles`.

Gacko:
### What does this PR do?

This PR makes `containerLogMaxSize` & `containerLogMaxFiles` of Kubelet configurable.

None. Both parameters are not being rendered by default. If they are set, they affect the log rotation behavior of Kubelet.

Towards https://github.com/giantswarm/roadmap/issues/3002.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -26,6 +26,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Add vertical-pod-autoscaler-crd HelmRelease (behind a flag which is disabled by default).
 - Add coredns HelmRelease (behind a flag which is disabled by default).
 - Support prepending cluster name to file secret name
+- Kubelet: Add `containerLogMaxSize` & `containerLogMaxFiles`. ([#92](https://github.com/giantswarm/cluster/pull/92))
 
 ### Changed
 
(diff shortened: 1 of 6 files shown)


njuettner (approved):

Gacko:
@njuettner Should we make the property public? There currently is no customizable Kubelet configuration in the "public" section, only below `internal`, so I wasn't sure wether I should really create a `kubelet` object in `providerIntegration`.

njuettner:
hmm I would like to leave that for now `internal` in case we migrate cluster, we should adjust it (meaning Giant Swarm) but not the customer.

Gacko:
Diff of the Kubelet configuration of two worker nodes, before and after changing `containerLogMaxSize` and `containerLogMaxFiles`:

```zsh
% diff ip-10-0-135-232.eu-west-2.compute.internal.yaml ip-10-0-239-28.eu-west-2.compute.internal.yaml 
102,103c102,103
<   containerLogMaxSize: 10Mi
<   containerLogMaxFiles: 5
---
>   containerLogMaxSize: 100Mi
>   containerLogMaxFiles: 10
```

The `kubelet-config` stored in the API server stays the same and customizations get applied node-locally through patching the `/var/lib/kubelet/config.yaml` upon joining.

PR #94, opened 2024-02-20
API Server: Make audit policy rules extendable.

Gacko:
### What does this PR do?

This PR makes the audit policy rules we already define extendable by customer defined rules.

The user can define additional audit policy rules which get added to the top of the list of default rules in the audit policy file.

Towards https://github.com/giantswarm/roadmap/issues/3002.


weseven (approved):

Gacko:
I've initially put the new key below `providerIntegration`, but according to the docs of that, I don't feel like this is correct. So I'll move it to a more common place.

PR #95, opened 2024-02-21
Put API server priority and fairness configuration behind a flag

nprokopic:
### What does this PR do?

This PR puts API server priority and fairness configuration behind a flag that is disabled by default.

API server priority and fairness config is not used by default.

### How does it look like?

See diff.

We have discovered issues with API server priority and fairness config which are breaking cluster creation and upgrades.


weseven (approved):

helm/cluster/templates/clusterapi/controlplane/_helpers_clusterconfiguration_apiserver.tpl
```diff
@@ -33,8 +35,10 @@ extraArgs:
   feature-gates: {{ include "cluster.internal.controlPlane.kubeadm.clusterConfiguration.apiServer.featureGates" $ }}
   {{- end }}
   kubelet-preferred-address-types: InternalIP
+  {{- if $.Values.internal.advancedConfiguration.controlPlane.apiServer.enablePriorityAndFairness }}
   max-mutating-requests-inflight: "${MAX_MUTATING_REQUESTS_INFLIGHT}"
```
calvix, line 39: what is the actual value and where does it come from?
is kubeadm somehow dynamically calculating this?

calvix (approved):

PR #99, opened 2024-02-22
downgrade k8s version to previous patch version

bdehri:
### What does this PR do?

This PR downgrades K8s version to previous patch release as we want to build a new flatcar image for current version without overriding existing clusters.

### How does it look like?

Towards https://github.com/giantswarm/giantswarm/issues/29892

Also see https://github.com/giantswarm/giantswarm/issues/29892#issuecomment-1958966726

No

Yes

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Changed
+
+- Downgrade Kubernetes version to v1.25.15.
+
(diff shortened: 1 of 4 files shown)


nprokopic:
Just testing, or we actually have to downgrade?

bdehri:
> Just testing, or we actually have to downgrade?

We want 1.25.16 k8s version to be on latest flatcar. As there is already a production CAPA cluster with 1.25.16, we rebuilt our flatcar image with latest flatcar, that cluster might get broken during ami update. long story short, we want to downgrade current prod capa cluster, rebuilt 1.25.16 with latest flatcar then upgrade the prod cluster.

PR #103, opened 2024-02-25
Add cluster chart docs

nprokopic:
Closes https://github.com/giantswarm/roadmap/issues/3195

### What does this PR do?

This PR adds cluster chart docs. This initial version of the docs mostly focuses on provider integration, i.e. the integration of the cluster chart into a cluster-\<provider\> app.

KaaS engineers can more easily work with cluster chart.

### How does it look like?

See the docs :)

https://github.com/giantswarm/roadmap/issues/3195

It's all about that here :)

No.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,3 +1,440 @@
-# Giant Swarm cluster helm chart
+# Cluster chart
 
-This repository contains all provider-independent resources for a Giant Swarm cluster (Cluster API clusters only, not for Vintage).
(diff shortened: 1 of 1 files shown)


nprokopic:
I will go ahead and merge this with my mighty admin privileges (😁), as it's the initial version of the docs that I promised and my vacation is starting tomorrow, but last week was hard, being sick and all that :) You can yell at me later 😬 🙈 ❤️

PR #107, opened 2024-02-27
Randomize etcd defragmentation start so they are staggered.

weseven:
### What does this PR do?

Adds a random, fixed delay to etcd defragmentation service start to prevent them being synchronized.

Towards https://github.com/giantswarm/giantswarm/issues/29868

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ## [Unreleased]
 
 - Apply API Server fairness settings using patches.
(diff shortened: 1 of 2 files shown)


njuettner (approved):

weseven:
> OOC do we know which version we run? According to the docs `RandomizedDelaySec` was added in version 247.
> 
> Do we know which version we run on our flatcar image 😅?

Oof, I tested this change on a cluster because I had the same doubt, but I forgot to check the version. Will check the version asap.
Anyway any flatcar stable image built later than March 3 2021 (flatcar 2765.2.0 or later) has systemd 247 or higher :)

PR #118, opened 2024-03-04
Chart: Make admission plugins configurable.

Gacko:
### What does this PR do?

This PR makes admission plugins configurable. This is required for porting the PodNodeSelector admission plugin from Vintage.

https://github.com/giantswarm/roadmap/issues/3002


njuettner (approved):
Lgtm, do we need to create the directory `/etc/kubernetes/admission/plugins/` or is it already present?

PR #126, opened 2024-03-14
Customize container registry and service account issuer via template and read provider specific helm values for coredns

calvix:
### What does this PR do?
changes for running `cluster-aws` in china

- coredns - add provider-specific helm values similar how it is done for cilium
- service account issuer - add option to generate the url via template
- container registry - add option to specify the registry either via raw value or via template

### What is the effect of this change on users?
It should be transparent, no need to change any value.
### What does it look like?

towards github.com/giantswarm/roadmap/issues/3257

related changes in `cluster-aws` - https://github.com/giantswarm/cluster-aws/pull/548


helm/cluster/README.md
```diff
@@ -490,6 +492,7 @@ Provider-specific properties that can be set by cluster-$provider chart in order
 | `providerIntegration.pauseProperties` | **Pause properties** - A map of property names and their values that will affect setting pause annotation|**Type:** `object`<br/>|
 | `providerIntegration.pauseProperties.*` |**None**|**Types:** `string, number, integer, boolean`<br/>|
 | `providerIntegration.provider` | **Provider** - The name of the Cluster API provider. The name here must match the name of the provider in cluster-<provider> app name.|**Type:** `string`<br/>|
+| `providerIntegration.registry` | **Registry** - The registry to use for pulling images. If not set, the default registry is used.||
```
njuettner, line 495: That looks weird, isn't the Type missing?

CHANGELOG.md
```diff
@@ -7,6 +7,12 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Added
+
+- Chart: Add `providerIntegration.controlPlane.kubeadmConfig.clusterConfiguration.apiServer.serviceAccountIssuer.templateName` to be able set API Service Account Issuer via template.
+- Chart: Add `providerIntegration.apps.corends` to beable to add provider specific coredns helm values.
```
njuettner, line 13:
```suggestion
- Chart: Add `providerIntegration.apps.corends` to be able to add provider specific coredns helm values.
```

njuettner, line 13:
```suggestion
- Chart: Add `providerIntegration.apps.coredns` to beable to add provider specific coredns helm values.
```

helm/cluster/README.md
calvix: could be becasue it can be null or object?

helm/cluster/values.schema.json
```diff
@@ -1955,6 +1976,51 @@
                     "title": "Provider",
                     "description": "The name of the Cluster API provider. The name here must match the name of the provider in cluster-<provider> app name."
                 },
+                "registry": {
+                    "title": "Registry",
+                    "description": "The registry to use for pulling images. If not set, the default registry is used.",
```
njuettner, line 1981: Just a question for my understanding, why don't we use "default" as a key (similar to what we use for "kubernetesVersion") to use the registry `gsoci.azurecr.io` instead of allowing one of `templateName` vs `value`?

njuettner: Or just setting the `gsoci.azurecr.io` registry in {{ define "cluster.image.registry" }}

calvix: So in our cluster-aws chart, we need to be able to provide the registry dynamically based on the region( atm based on `cn-` prefix in the region) so the customer or user doesn't need to provide a bunch of values for the cluster
 
 The only way how to do that is by using a template and providing the template name
, I added Also the `value` as this seems to be the common option that is used in `cluster` chart so wanted to keep that similar, but it is not used nor needed for our changes, so I would be fine to drop it. 

I didn't want to provide another default as the registry is internally already defined under `.values.internal.advancedConfiguration.registry` and so the changes either use what is set by the default there or use the value provided in `.providerIntegration`

calvix: >  setting the gsoci.azurecr.io registry in {{ define "cluster.image.registry" }}


That was also in my head but I did not like that a default is hidden in a function/template somewhere

calvix: Anyway what whatever makes you happy I can do, This just felt the right approach but I do not have that much experience with the `cluster` chart so it might not be the best, feel free to suggest better solution

Gacko:
As much as I appreciate your grade of automation, but could you please give the PR a descriptive summary? Thanks!

helm/cluster/templates/_helpers.tpl
```diff
@@ -141,3 +141,18 @@ Where `data` is the data to hash and `global` is the top level scope.
 {{- $testValue := .testValue }}
 {{ eq $propertyValue $testValue }}
 {{- end }}
+
+
```
Gacko, line 145:
```suggestion
```

helm/cluster/templates/_helpers.tpl
```diff
@@ -141,3 +141,18 @@ Where `data` is the data to hash and `global` is the top level scope.
...
+{{/* Function to determine the value of container image registry that is used across whole repository   */}}
```
Gacko, line 146:
```suggestion
{{/* Function to determine the value of container image registry that is used across whole repository */}}
```

helm/cluster/templates/_helpers.tpl
```diff
@@ -141,3 +141,18 @@ Where `data` is the data to hash and `global` is the top level scope.
...
+{{- if $.Values.providerIntegration.registry -}}
+{{- if $.Values.providerIntegration.registry.templateName -}}
+{{- $registry = ( include $.Values.providerIntegration.registry.templateName $ ) -}}
+{{- else if $.Values.providerIntegration.registry.value -}}
+{{- $registry = $.Values.providerIntegration.registry.value -}}
+{{- end -}}
+{{- end -}}
+{{- $registry -}}
+{{- end -}}
+
```
Gacko, line 158:
```suggestion
```

helm/cluster/values.schema.json
njuettner: If we don't use the value I'd say lets drop it and leave the rest 👍🏻

calvix: ACK , dropping

helm/cluster/README.md
calvix: fixed

helm/cluster/values.schema.json
calvix: Done

calvix:
> As much as I appreciate your grade of automation, but could you please give the PR a descriptive summary? Thanks!

Do you mean the PR title? I changed that to something more specific, Or do you mean the the rest of the text? What are you missing there,   For my head it kinda works :sweat_smile:

njuettner (approved):
lgtm from my side

Gacko:
> Do you mean the PR title?

Yeah, just the title. This will end up as the commit message once merged and so it would be nice to have it a bit more "readable". Thanks for changing! ❤️

PR #127, opened 2024-03-14
Add default apps that we use in CAPA at the moment

nprokopic:
Closes https://github.com/giantswarm/roadmap/issues/3299

Towards https://github.com/giantswarm/roadmap/issues/3119

#### TL;DR

This PR adds default apps that we use in CAPA at the moment.

It moves apps from default-apps-aws into cluster chart and focuses on:
- Moving the apps we need
- Preserving the old functionality
- Matching App CR config to look like HelmRelease config
- keeping required changes in capi-migration-cli to a minimum 

#### Added

- Add capi-node-labeler app (disabled by default).
- Add cert-exporter app (disabled by default).
- Add cert-manager app (disabled by default).
- Add chart-operator-extensions app (disabled by default).
- Add cilium-servicemonitors app (disabled by default).
- Add cluster-autoscaler app (disabled by default).
- Add etcd-kubernetes-resources-count-exporter app (disabled by default).
- Add external-dns app (disabled by default).
- Add k8s-dns-node-cache app (disabled by default).
- Add metrics-server app (disabled by default).
- Add net-exporter app (disabled by default).
- Add node-exporter app (disabled by default).
- Add observability-bundle app (disabled by default).
- Add security-bundle app (disabled by default).
- Add teleport-kube-agent app (disabled by default).
- Add vertical-pod-autoscaler app (disabled by default).
- Add `$.Values.providerIntegration.apps.cilium.enable` flag to enable Cilium HelmRelease (old flag `$.Values.providerIntegration.resourcesApi.ciliumHelmReleaseResourceEnabled` is deprecated).
- Add `$.Values.providerIntegration.apps.coreDns.enable` flag to enable CoreDns HelmRelease (old flag `$.Values.providerIntegration.resourcesApi.coreDnsHelmReleaseResourceEnabled` is deprecated).
- Add `$.Values.providerIntegration.apps.networkPolicies.enable` flag to enable Network policies HelmRelease (old flag `$.Values.providerIntegration.resourcesApi.networkPoliciesHelmReleaseResourceEnabled` is deprecated).
- Add `$.Values.providerIntegration.apps.verticalPodAutoscalerCrd.enable` flag to enable Network policies HelmRelease (old flag `$.Values.providerIntegration.resourcesApi.verticalPodAutoscalerCrdHelmReleaseResourceEnabled` is deprecated).
- Add `$.Values.internal.ephemeralConfiguration.apps` config, meant only for development and temporary problem mitigation purposes, and where version and catalog can be overridden for every app.

#### Fixed

- Fix CoreDNS provider-specific config (it was incorrectly reading Cilium app config instead of CoreDNS app config).

For users who create the cluster - with this change and the accompanying change in cluster-aws, now they apply a single App when creating a cluster, only cluster-aws, as opposed to previously having to apply cluster-aws and default-apps-aws.

For Giant Swarm KaaS engineers who work on the CAPA product - changes for provider-independent apps are done in the cluster chart, and not in defaulta-apps-aws.

### How does it look like?

There are changes in how apps will be configured, considering that they are not in a different chart. There are also changes in how we define which default apps we have, where some of things that were (in theory) configurable via Helm values before are not configurable like that anymore (e.g. app versions, or being able to add a whole new app).

#### App config changes

Previously the app config was set via default-apps-aws Helm values like this:

```yaml
userConfig:
  foo-bar:
    configMap:
      values:
        someValue: 123
        otherValue: abc
    secret:
      <unencoded plain text string data>
  awesomeExporter:
    configMap:
      values:
        hello: world
    secret:
      <unencoded plain text string data>
```

Now the above becomes the following config of cluster chart (and therefore of cluster-aws):

```yaml
global:
  apps:
    fooBar:
      values:
        someValue: 123
        otherValue: abc
    awesomeExporter:
      values:
        hello: world
```

Changes that happen here, besides the place where we have these Helm values:
- Case of Helm property names: previously they were camelCase in some cases, and hyphen-string-case in some cases. Now they are all camelCase, like everything else in cluster chart Helm values.
- Secret config that there was in default-apps-aws at `.Values.userConfig.<APP>.secret` is dropped for the following reasons:
  - It's not used anywhere in the default-app-aws by default for the current apps.
  - capi-migration-cli is not using this for the migration anywhere.
  - Putting a secret data there means that the data that should supposedly be a "secret" will end up in a ConfigMap as a plain text (in user-values ConfigMap), so there no value in doing it and it is even misleading as somebody might expect that they are securely setting some secret values there.
  - If some specific cluster is using `.Values.userConfig.<APP>.secret` from default-apps-aws (if anyone then it's maybe us in our CAPA MCs/WCs), then they can apply that secret via `.Values.global.<app>.extraConfigs` when doing the upgrade to the new version (manual changes will be required in any case for the all other app configs as well).

#### App list and spec changes

Previously the apps were specified in default-apps-aws Helm values (in values.yaml) like this:

```yaml
apps:
  foo-bar:
    appName: foo-bar
    chartName: foo-bar
    catalog: default
    clusterValues:
      configMap: true
      secret: false
    dependsOn: baz-app
    enabled: true
    namespace: kube-system
    # used by renovate
    # repo: giantswarm/foo-bar
    version: 1.14.1
  awesomeExporter:
    appName: awesome-exporter
    chartName: awesome-exporter
    catalog: default
    clusterValues:
      configMap: true
      secret: true
    enabled: true
    forceUpgrade: true
    namespace: kube-system
    # used by renovate
    # repo: giantswarm/awesome-exporter
    version: 0.7.1
    # a list of extraConfigs for the App,
    # It can be secret or configmap
    # https://github.com/giantswarm/rfc/tree/main/multi-layer-app-config#example
    extraConfigs:
      - kind: configMap
        name: "{{ $.Values.clusterName }}-awesome-exporter-config"
        namespace: "{{ $.Release.Namespace }}"
```

The above approach exposes the entire list of apps and all details via Helm values. Most of those values should not be configurable, but configurability of some of those may be needed during the development and testing.

So in the cluster chart, the above app list Helm values have been moved to `helm/cluster/files/apps`, where there is a file per app now.

So the above now becomes:

```yaml
# helm/cluster/files/apps/foo-bar.yaml
appName: foo-bar
chartName: foo-bar
configKey: fooBar
catalog: default
clusterValues:
  configMap: true
  secret: false
dependsOn: baz-app
namespace: kube-system
# used by renovate
# repo: giantswarm/foo-bar
version: 1.14.1
```

and

```yaml
# helm/cluster/files/apps/awesome-exporter.yaml
appName: awesome-exporter
chartName: awesome-exporter
configKey: awesomeExporter
catalog: default
clusterValues:
  configMap: true
  secret: true
forceUpgrade: true
namespace: kube-system
# used by renovate
# repo: giantswarm/awesome-exporter
version: 0.7.1
# a list of extraConfigs for the App,
# It can be secret or configmap
# https://github.com/giantswarm/rfc/tree/main/multi-layer-app-config#example
extraConfigs:
  - kind: configMap
    name: "{{ $.Values.global.metadata.name }}-awesome-exporter-config"
    namespace: "{{ $.Release.Namespace }}"
```

With `extraConfigs` being configurable in the public API mentioned above, the entire new public config is now like this:

```yaml
global:
  apps:
    fooBar:
      values:
        someValue: 123
        otherValue: abc
      extraConfigs:
      - name: my-secret-extra-config-1
        kind: Secret
        priority: 110
    awesomeExporter:
      values:
        hello: world
      - name: my-extra-config-1
        kind: ConfigMap
        priority: 110
```

The above shows how we configure apps that we deploy as App CRs, which is now almost\* the same as the API we use for configuring apps that we deploy as HelmRelease. So when we switch all apps to use one approach, it will not be a breaking change in most of the cases.

\* Almost the same because App ExtraConfig has `priority`, while HelmRelease ExtraConfig does not have that. Also HelmRelease ExtraConfig has `optional`, and App ExtraConfig does not.

Also we now have the following ephemeral config that can be used for the purposes of development, testing and putting out a production fire in the middle of the night:

```yaml
internal:
  ephemeralConfiguration:
    apps:
      fooBar:
        versionOverride: v100.0.0-dev.8234y65dh92783
        catalogOverride: default-test
        disable: false
```

Notes:
- Previously we had `.Values.<app>.enabled`, and now it's `.Values.internal.advancedConfiguration.<app>.disable` so we can keep default false value (and have all apps enabled by default), and not pollute default Helm values.
- Previously we had `.Values.<app>.version`, now it's more explicit name `.Values.internal.ephemeralConfiguration.<app>.versionOverride`
- Previously we had `.Values.<app>.catalog`, now it's more explicit name `.Values.internal.ephemeralConfiguration.<app>.catalogOverride`
- Other old `.Values.<app>.*` Helm values, like app version, dependencies, etc, are not configurable anymore.
- This app config is under `.Values.internal.ephemeralConfiguration` so even more explicit about the purpose and (lack of) durability of the values set there.

- https://github.com/giantswarm/roadmap/issues/3299
- https://github.com/giantswarm/roadmap/issues/3119

Check this out and try to understand if this works for us, or if I'm missing something here.

### Upgrade and migration

3-step upgrade process (with simplified details for the sake of brevity):

1. **Upgrade default-apps-aws version**
  - Apps get paused (so that Charts on the WC are not deleted)
  - new Helm hooks are created (post-upgrade hook to propagate pause to bundled apps, post-delete hook to remove App CRs that stay behind due to finalizers, which happens because of above pause)
2. **Delete default-apps-aws**
  - All App CRs are deleted
  - Chart CRs stay on WC and apps are still running
3. **Upgrade cluster-aws to a new version with default apps**
  - cluster upgraded successfully
  - All new App CRs for default apps got created
  - Chart CRs (for default apps) in the WC got updated to reflect that they are now deployed by cluster-aws

Yes. Will do that as well.


AndiDog:
Looks great. I'd rename `s/disabled/disable/` because that's consistent with _most_ Helm charts.

njuettner:
If you cannot change the version of the apps, how do you ensure testing?

helm/cluster/values.schema.json
```diff
@@ -39,22 +39,42 @@
...
                             },
-                            "optional": {
-                                "type": "boolean",
-                                "title": "Optional",
-                                "description": "Optional marks this ValuesReference as optional. When set, a not found error for the values reference is ignored, but any ValuesKey, TargetPath or transient error will still result in a reconciliation failure."
+                            "priority": {
+                                "type": "integer",
+                                "title": "Priority",
+                                "default": 25,
+                                "max": 150,
```
njuettner, line 46: Why do you have max 150? Is this arbitrary?

nprokopic:
> Looks great. I'd rename `s/disabled/disable/` because that's consistent with _most_ Helm charts.

Thanks! Thanks for the tip, I have renamed `disabled` to `disable`.

> If you cannot change the version of the apps, how do you ensure testing?

This was an oversight, fixed it and added `versionOverride` property.

nprokopic:
I have made another change to the initial PR version, "internal" app config that was initially here:

```yaml
internal:
  advancedConfiguration:
    apps:
      fooBar: {}
      myExporter: {}
```

since the use case for the config there is about the development, testing and extreme cases of operational work where there is some fire that has to be put out quickly, and all those configs are something that should be temporary and not be persisted (for a long time), I have moved those to the new `.internal.ephemeralConfiguration` where the description is even more explicit and concrete (compared to `internal.advancedConfiguration`, so now it looks like this:

```yaml
internal:
  ephemeralConfiguration:
    apps:
      fooBar:
        version: 100.0.0-dev.8234y65dh92783
        catalog: default-test
      myExporter:
        disable: true
```

In the JSON schema the description of the ephemeral configuration is:
> This configuration is meant for Giant Swarm staff only, changes here are ephemeral and are not meant to be persistent and it can be removed at any point. It can be used for the purposes of development, testing and operations.

And the description of the app version there is:
> Custom application version that overrides the application version from the release. This is usually a new development version that you want to test, or a newer patch version that you need to deploy in order to put out a production fire in the middle of the night. Use carefully!

helm/cluster/files/apps/external-dns.yaml
```diff
@@ -0,0 +1,21 @@
...
+dependsOn: prometheus-operator-crd
+enabled: true
+namespace: kube-system
+# used by renovate
+# repo: giantswarm/external-dns-app
+version: 3.1.0
+defaultValues:
+  txtOwnerId: giantswarm-io-external-dns
+  txtPrefix: {{ $.Values.global.metadata.name }}
+  annotationFilter: giantswarm.io/external-dns=managed
```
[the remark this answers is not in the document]
nprokopic: Thanks! Yeah, I didn't add this yet, since this named template is not added yet, and I needed the rendering to work locally while I was doing the work. Will add it.

helm/cluster/values.schema.json
nprokopic: According to our [docs](https://docs.giantswarm.io/vintage/getting-started/app-platform/app-configuration/#extra-configs) the max priority of an extraConfig is 150.

helm/cluster/files/apps/external-dns.yaml
```diff
@@ -0,0 +1,20 @@
...
+  configMap: true
+  secret: true
+dependsOn: prometheus-operator-crd
+namespace: kube-system
+# used by renovate
+# repo: giantswarm/external-dns-app
+version: 3.1.0
+defaultValues:
+  txtOwnerId: giantswarm-io-external-dns
+  txtPrefix: {{ $.Values.global.metadata.name }}
```
mcharriere, line 15: we added this new value today which is needed for some cases. I'd be nice to include it here already.

```suggestion
  txtPrefix: {{ $.Values.global.metadata.name }}
  registry: txt
```

calvix:
How far are we with merging @nprokopic ?

nprokopic:
> How far are we with merging @nprokopic ?

Sorry, it got delayed due to KubeCon last week. I picked it up again this week, so hopefully it's done in the following few days.

helm/cluster/files/apps/external-dns.yaml
nprokopic: Do we need this for all providers?

I'm working ATM on enabling provider-specific config for apps, so every cluster-\<provider\> app can set provider-specific config it needs for that app (provider-specific config will be merged with and will overwrite provider-independent default values here).

mcharriere: yes, for all providers is fine. It's just a safeguard, because there is a value in the catalog overriding our default in the app.

nprokopic: Thanks, added it.

helm/cluster/templates/apps/apps.yaml
```diff
@@ -0,0 +1,137 @@
...
+kind: ConfigMap
+metadata:
+  labels:
+    {{- include "cluster.labels.common.all" $ | nindent 4 }}
+  name: {{ $.Values.global.metadata.name }}-{{ $appName }}-user-values
+  namespace: {{ $.Release.Namespace }}
+data:
+  values: |
+  {{- (tpl ($userValues | toYaml | toString) $) | nindent 4 }}
+{{- /*end if $hasUserValues*/}}
```
Gacko, line 129:
```suggestion
{{- /* end if $hasUserValues */ }}
```

helm/cluster/templates/apps/apps.yaml
```diff
@@ -0,0 +1,137 @@
...
+{{- end }}
+
+{{- /* end if not $ephemeralConfig.disable*/}}
```
Gacko, line 132:
```suggestion
{{- /* end if not $ephemeralConfig.disable */ }}
```

helm/cluster/templates/apps/apps.yaml
```diff
@@ -0,0 +1,137 @@
...
+{{- end }}
+{{- /* end with $rootScope */}}
```
Gacko, line 134:
```suggestion
{{- /* end with $rootScope */ }}
```

helm/cluster/templates/apps/apps.yaml
```diff
@@ -0,0 +1,137 @@
...
+{{- end }}
+{{- /* end outer range */}}
```
Gacko, line 136:
```suggestion
{{- /* end outer range */ }}
```

Gacko:
Just a dumb question as I'm just stumbling across that while migrating `aws-pod-identity-webhook` to `cluster-aws`:

We're referencing `<cluster>-cluster-values` in the config section of the created apps. This ConfigMap gets created by `cluster-apps-operator` upon creation of Cluster CRs. Also we are referencing the `<cluster>-kubeconfig` which gets created just after creating the control plane.

Unfortunately the app operator's webhook does reject apps for which either the `<cluster>-cluster-values` or `<cluster>-kubeconfig` are missing. How do we overcome that?

Gacko:
https://github.com/giantswarm/cluster/pull/127/files#diff-f52d66a3733fb9178e02dedf330e3ae86fc6d472008fe088f935217c8963e687R33-R34

That needs to be `configMap` and `secret` for apps. I know, that sucks, just ran into that issue. For `HelmReleases` it's indeed `ConfigMap` and `Secret`.

Gacko:
@nprokopic: I updated the `app` schema to match the actual API: https://github.com/giantswarm/apiextensions-application/blob/main/api/v1alpha1/app_types.go#L135

helm/cluster/templates/apps/apps.yaml
```diff
@@ -0,0 +1,141 @@
+{{- $rootScope := . }}
+{{- range $path, $_ := .Files.Glob "files/apps/**.yaml" }}
+{{- $app := dict }}
+{{- with $rootScope }}
+{{- $app = (tpl (.Files.Get $path) $) | fromYaml }}
+{{- $appName := $app.appName }}
+{{- $appKey := $app.configKey }}
```
AverageMarcus, line 7: Do we need a separate `configKey`? Could we instead use `appName` as we can then easily work this out in testing from an apps own PR (for example).
If I understand correctly the only reason for using a `configKey` is to allow the name to not have hyphens in it but as we are only using it with the `get` calls below it shouldn't make a difference, right?

Otherwise, we need some way of building the `configMap` based on the app repo.

helm/cluster/values.yaml
```diff
@@ -68,11 +84,49 @@ internal:
         memory: 384Mi
     registry: gsoci.azurecr.io
     workers: {}
+  ephemeralConfiguration: {}
```
AverageMarcus, line 87: Can we also have similar to this for `HelmReleases` (e.g. Cilium) so that we can override their catalog and version in testing too.

nprokopic: Good point, thanks for bringing it up. We should add that definitely! (just not sure if in this PR, as it's already really big change)

AverageMarcus: Yeah no worries about it being a different PR. Just wanted to mention it somewhere as I went through this PR with Tinkerers and Turtles yesterday thinking about how we're going to go about testing default apps. 😁 Was very happy to find it already considered.

helm/cluster/templates/apps/apps.yaml
nprokopic: `configKey` is the name of the app config entry in these Helm values:
- `.Values.global.apps`,
- `.Values.providerIntegration.apps` and in `.Values.internal.ephemeralConfiguration.apps`

so the key is meant to be predefined/known API with naming aligned with all other Helm values (camelCase here), especially for `.Values.global.apps` which a public API for customers.

> Otherwise, we need some way of building the configMap based on the app repo.

Where and why you need this?

Do you have some case where you dynamically/programatically build app configs?

Gacko (approved):

helm/cluster/values.yaml
AverageMarcus: @nprokopic Any objection to me creating an issue for this specific requirement (overriding HelmRelease version) so that I can track and reference it from our issues?

nprokopic:
Cluster that was created with apps in default-apps-aws successfully upgraded to a cluster with default apps in cluster-aws :tada:

3-step upgrade process (with simplified details for the sake of brevity):

1. **Upgrade default-apps-aws version**
  - Apps get paused (so that Charts on the WC are not deleted) :white_check_mark:
  - new Helm hooks are created (post-upgrade hook to propagate pause to bundled apps, post-delete hook to remove App CRs that stay behind due to finalizers, which happens because of above pause) :white_check_mark:
2. **Delete default-apps-aws**
  - All App CRs are deleted :white_check_mark:
  - Chart CRs stay on WC and apps are still running :white_check_mark:
3. **Upgrade cluster-aws to a new version with default apps**
  - cluster upgraded successfully :white_check_mark:
  - All new App CRs for default apps got created :white_check_mark:
  - Chart CRs (for default apps) in the WC got updated to reflect that they are now deployed by cluster-aws :white_check_mark:

I would merge these PRs now (after re-running cluster-test-suites, which I think I would have to do locally, due to nature of the breaking changes here):
- https://github.com/giantswarm/cluster/pull/127
- https://github.com/giantswarm/cluster-aws/pull/581
- https://github.com/giantswarm/default-apps-aws/pull/463

Other tools/components (I can remember now) that need changes are then (should not be a lot of work):
- capi-migration-cli
- kubectl-gs
- happa
- cluster-test-suites (upgrade suite will be failing until there is a new cluster-aws release)

P.S. I would probably merge the above PRs even without changes to mentioned tools/components, so we can upgrade clusters ASAP.

helm/cluster/values.yaml
nprokopic: Sorry for the late response, I was on a vacation. Feel free to create an issue (if you haven't already). Thanks!

AverageMarcus: I had actually forgot but this was a good reminder :D Here you go - https://github.com/giantswarm/roadmap/issues/3440

PR #152, opened 2024-04-24
Add an ability to use non-https registries

allanger:
In order to make the "per WC" cache work, we need to use http://127.0.0.1 in a containerd config, because the caching registry is exposed as a node port.

It would also be possible to use tls, but then we need to have a cert that is trusted by containerd, and it doesn's sount that easy

Issue: https://github.com/giantswarm/giantswarm/issues/30596

### What does this PR do?

This PR is supposed to let use http registries mirrors, when it's explicitly provided via endpoints

Currently, when I try to set `http://127.0.0.1:5000` as a mirror, I'm getting a config that contains `https://http://127.0.0.1`.

It should not affect users. 

It would make `per WC` cache deployment a way easier

Check, if this change doesn't affect users unless they explicitly want to be affected

I don't think so

I don't think so


njuettner:
@giantswarm/team-turtles what about setting this as a hardcoded string? Meaning adding "http://127.0.0.1" directly underneath the range?

That would at least prevent adding more non-https value endpoints. So we basically don't add additional endpoints via https and http.

Thoughts?

allanger:
> what about setting this as a hardcoded string?

We won't be able to hardcode the port though, would you be ok with something like that:

```yaml
#values
localMirror:
  enable: true
  port: 33445
```

```
# template
...
{{ - if .Values.localMirror.enabled -}}
"http://127.0.0.1:{{ .Values.localMirror.port }}
{{- end -}}
...
```

allanger:
But with that values structure, I'm a bit uncertain where to put that `localCache` entry and what it should look like when creating a cluster via happa, do you have ideas?

allanger:
@giantswarm/team-turtles  Is it better now?

allanger:
@weseven or @njuettner 
May I please ask you to review it again?

weseven:
LGTM, only nitpicks would be naming of values:
 - instead of `localCache` I would use `localRegistryCache`
 - instead of `mirrors` I'd use `mirroredRegistries`
 But I don't feel strongly about these, so not blocking.
 
 Not sure if .internal it's the right place: I think so, but better wait for @njuettner confirmation on that before merging.

weseven (approved):

njuettner:
Agree it makes sense to name them as Daniel suggested
localCache ->localRegistryCache
mirrors -> mirroredRegistries

to make it more explicit 🙂 

Other than that LGTM, feel free merge it once you changed the names

PR #166, opened 2024-05-01
Enable cloud-config

njuettner:
In order to integrate Azure as a provider we must integrate cloud-config which is used by several components (Azure CSI, Azure Cloud Controller Manager, API, ...)

In order to make it flexible I added a new cloud config value for the path `e.g. cloud-config: /etc/kubernetes/azure.json`, provider chart can set a default if needed otherwise it will be ignored.

I have opened a PR for cluster-azure which shows how it's used: https://github.com/giantswarm/cluster-azure/pull/264


helm/cluster/templates/clusterapi/controlplane/_helpers_files.tpl
```diff
@@ -7,6 +7,19 @@
...
+{{- include "cluster.internal.controlPlane.kubeadm.files.cloudConfig" $ }}
+{{- end }}
+
+{{- define "cluster.internal.controlPlane.kubeadm.files.cloudConfig" }}
+{{- if $.Values.providerIntegration.controlPlane.kubeadmConfig.clusterConfiguration.apiServer.cloudConfig  }}
+- path:  {{ $.Values.providerIntegration.controlPlane.kubeadmConfig.clusterConfiguration.apiServer.cloudConfig  }}
+  permissions: "0644"
+  contentFrom:
+    secret:
+      name: {{ include "cluster.resource.name" $ }}-control-plane-{{ include "cluster.data.hash" (dict "data" (include $.Values.providerIntegration.controlPlane.resources.infrastructureMachineTemplateSpecTemplateName $) "salt" $.Values.providerIntegration.hashSalt) }}-{{ $.Values.providerIntegration.provider }}-json
```
njuettner, line 19: because we are using a hash for CP templates we also need that.

IIRC `CAPZ controller creates the secret when oberserving the applied CR's, so it uses inside the name of the secret the hash.

Gacko (approved):

weseven (approved):

PR #170, opened 2024-05-06
Allow to set SELinux mode through `global.components.selinux.mode`.

weseven:
### What does this PR do?

Towards: https://github.com/giantswarm/giantswarm/issues/30605

Allows to set SELinux mode (enforcing, permissive or disabled) through `global.components.selinux.mode`.
Default value is `permissive` (current flatcar default).

- [x] CHANGELOG.md has been updated

### DIFF
--- a/helm/cluster/files/etc/selinux/config
+++ b/helm/cluster/files/etc/selinux/config
@@ -1,0 +1,15 @@
+# This file controls the state of SELinux on the system on boot.
+
+# SELINUX can take one of these three values:
+#       enforcing - SELinux security policy is enforced.
+#       permissive - SELinux prints warnings instead of enforcing.
+#       disabled - No SELinux policy is loaded.
+SELINUX={{ $.Values.global.components.selinux.mode }}
+
+# SELINUXTYPE can take one of these four values:
+#       targeted - Only targeted network daemons are protected.
+#       strict   - Full SELinux protection.
+#       mls      - Full SELinux protection with Multi-Level Security
+#       mcs      - Full SELinux protection with Multi-Category Security
+#                  (mls, but only one sensitivity level)
+SELINUXTYPE=mcs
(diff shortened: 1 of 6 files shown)


weseven:
@stone-z Please take a look, we will be able to set SELinux to enforcing, permissive (current default) or disabled through helm values (once this is merged and cluster-$provider charts pick up the new version and update their schema, e.g., https://github.com/giantswarm/cluster-aws/pull/610).

Do we also want to allow to set the type of protection (targeted, strict, mls, mcs) through values?
(Changes will be reflected in [the /etc/selinux/config file](https://github.com/giantswarm/cluster/blob/0077ba73de5cb818f17932f4d0b6f423e189e909/helm/cluster/files/etc/selinux/config))

helm/cluster/files/etc/selinux/config
```diff
@@ -0,0 +1,15 @@
...
+#       disabled - No SELinux policy is loaded.
+SELINUX={{ $.Values.global.components.selinux.mode }}
+
+# SELINUXTYPE can take one of these four values:
+#       targeted - Only targeted network daemons are protected.
+#       strict   - Full SELinux protection.
+#       mls      - Full SELinux protection with Multi-Level Security
+#       mcs      - Full SELinux protection with Multi-Category Security
+#                  (mls, but only one sensitivity level)
+SELINUXTYPE=mcs
```
stone-z, line 15: > Do we also want to allow to set the type of protection (targeted, strict, mls, mcs) through values?

If `mcs` is the current default, do you see any clear situations where customers would want to change the mode?

IIUC, MCS is actually an extension of MLS, and both supplement `strict`. So in theory, MCS would be the most fully featured option from an enforcement perspective. Is that correct?

stone-z (approved):

helm/cluster/files/etc/selinux/config
weseven: 100%, honestly I don't think customers would want to change it, and I don't think we would want to test it in any other way, so I think we're fine just giving the option to set enforcing/permissive or disable SELinux.

stone-z: That's my impression as well. Then this looks great :+1:

PR #183, opened 2024-05-14
Add MachineHealthCheck and MachineDeployment resource template to NodePools.

njuettner:
Towards: https://github.com/giantswarm/roadmap/issues/3121

### What does this PR do?

A couple of changes:

- changing `COREOS_EC2*` variables to support multiple providers (HOSTNAME, IPV4_LOCAL needs to be replaced by each provider, e.g. via preKubeadmCommands) 
- add MachineDeployment and MachineHealth for NodePools to support Azure
- add additional schema field infrastructureMachineTemplateSpecTemplateName to calculate hash for MachineDeployments
- move healthCheck object to definitions because it's now used by ControlPlane and MachineDeployments


njuettner:
The reason of the diff (2 files removed) is due to the name has now changed of the machinepools because of the hash ending since we now using HOSTNAME and IPV4_LOCAL instead of COREOS_EC2* variables

weseven (approved):

helm/cluster/values.schema.json
```diff
@@ -2329,6 +2337,23 @@
                         }
                     }
                 },
+                "environments": {
```
nprokopic, line 2340: small nit, not that important, I would call it `environmentVariables`

nprokopic (approved):
LGTM

.github/workflows/zz_generated.diff_helm_render_templates.yaml
```diff
@@ -89,7 +89,7 @@ jobs:
         run: |
           helm dependency build old/helm/${{ github.event.repository.name }}
 
-          for test_file_path in helm/${{ github.event.repository.name }}/ci/test-*-values.yaml; do
+          for test_file_path in old/helm/${{ github.event.repository.name }}/ci/test-*-values.yaml; do
```
Gacko, line 92: This is unintentional, right?

njuettner: Yeah you're right, will clean it up 👍🏻

PR #201, opened 2024-05-28
API Server: Make feature gates configurable.

Gacko:
### What does this PR do?

This PR makes the list of feature gates passed to the API server configurable not only via provider integration but also via values passed to the cluster chart by users.

### What is the effect of this change to users?

Users now can configure features gates on their own and do not need to depend on the provider specific cluster chart's defaults.

### Any background context you can provide?

Closes https://github.com/giantswarm/roadmap/issues/3333.

### Should this change be mentioned in the release notes?

- [x] CHANGELOG.md has been updated (if it exists)

PR #202, opened 2024-05-29
API Server: Fix feature flag condition.

Gacko:
Also improve order of additional admission plugins.

### What does this PR do?

This PR makes the feature flag condition also check for `cluster` chart defined feature flags. Additionally it correctly orders the way additional admission plugins are being concatenated, even though this doesn't change the output because they are not overriding each other - at least not yet.

### How does it look like?

Follow-up of https://github.com/giantswarm/cluster/pull/201.


njuettner (approved):
LGTM, don't forget to add CHANGELOG entry

Gacko:
Actually the changelog was already added in #201, this is just a follow up.

PR #211, opened 2024-06-04
Add support for Release CRs

nprokopic:
### Added

This PR adds support for using Release CRs and getting Kubernetes, Flatcar and apps version, catalog and dependencies from there.

The change is fully backward compatible, and it requires that new flag `$.Values.providerIntegration.useReleases` is set to `true`. The default value of `$.Values.providerIntegration.useReleases` is `false`, so cluster chart version can be
safely updated.

See https://github.com/giantswarm/rfc/pull/96 for more details.

https://github.com/giantswarm/rfc/pull/96


helm/cluster/templates/_helpers.tpl
```diff
@@ -151,4 +151,115 @@ Where `data` is the data to hash and `global` is the top level scope.
 {{- end -}}
 {{- end -}}
 {{- $registry -}}
-{{- end -}}
\ No newline at end of file
+{{- end -}}
+
+{{- define "cluster.internal.get-provider-integration-values" }}
```
nprokopic, line 156: Will add comments that explain this, but briefly...

Helm templates have a different context, depending on how you call them, from the chart where the template is defined, or from the parent chart for example.

E.g. in cluster chart we have `$.Values.providerIntegration`, if you use it in a template defined in cluster chart and you call that template from the cluster chart as well, then you just use `$.Values.providerIntegration` in the template.

OTOH if you call that same template from a parent chart (e.g. from cluster-aws), then that same Helm value is in `$.Values.cluster.providerIntegration`.

So this helper template here check from where the call has been made, reads either `$.GiantSwarm.providerIntegration` or `$.GiantSwarm.cluster.providerIntegration` and sets that to `$.GiantSwarm.providerIntegration`.

This way we can use `$.GiantSwarm.providerIntegration` in all templates (after calling this helper).

This also simplifies a lot how other shared templates are called, so it's simpler to use them from e.g. cluster-aws and other charts.

helm/cluster/templates/_helpers.tpl
```diff
@@ -151,4 +151,115 @@ Where `data` is the data to hash and `global` is the top level scope.
...
+{{- end }}
+{{- /* Create $.GiantSwarm object where we put custom Giant Swarm vars */}}
+{{- if not $.GiantSwarm }}
+  {{- $_ := set $ "GiantSwarm" dict }}
+{{- end }}
+{{- /* Finally set "UseReleases" property that can be used in other templates */}}
+{{- $_ := set $.GiantSwarm "providerIntegration" $providerIntegration }}
+{{- end }}
+
+{{- define "cluster.internal.get-internal-values" }}
```
nprokopic, line 173: Same like above for `cluster.internal.get-provider-integration-values` template.

helm/cluster/templates/_helpers.tpl
```diff
@@ -151,4 +151,115 @@ Where `data` is the data to hash and `global` is the top level scope.
...
+{{- range $_, $app := $.GiantSwarm.Release.spec.apps }}
+{{- if eq $app.name $.appName }}
+{{- $appVersion = $app.version }}
+{{- end }}
+{{- end }}
+{{- end }}
+{{- $appVersion }}
+{{- end }}
+
+{{- define "cluster.component.version" }}
```
nprokopic, line 216: Similar like for `cluster.app.version`, just for components (e.g. Kubernetes and OS version).

helm/cluster/templates/_helpers.tpl
```diff
@@ -151,4 +151,115 @@ Where `data` is the data to hash and `global` is the top level scope.
...
+{{- range $_, $component := $.GiantSwarm.Release.spec.components }}
+{{- if eq $component.name $.componentName }}
+{{- $componentVersion = $component.version }}
+{{- end }}
+{{- end }}
+{{- end }}
+{{- $componentVersion }}
+{{- end }}
+
+{{- define "cluster.component.kubernetes.version" }}
```
nprokopic, line 229: Helper to get Kubernetes version:
- For providers that use Release CRs, it gets k8s version from the release CR.
- For providers that do not use Release CRs, it gets k8s version from the Helm value like before.

This way we can release cluster chart and during the transition phase we can use the same cluster chart version across different providers, where some providers use new releases and others do not.

helm/cluster/templates/_helpers.tpl
```diff
@@ -151,4 +151,115 @@ Where `data` is the data to hash and `global` is the top level scope.
...
+{{- $_ := set $ "componentName" "kubernetes" }}
+{{- include "cluster.component.version" $ | trimPrefix "v" }}
+{{- else if $.GiantSwarm.providerIntegration.kubernetesVersion }}
+{{- $.GiantSwarm.providerIntegration.kubernetesVersion | trimPrefix "v" }}
+{{- else }}
+{{- fail "Cannot determine Kubernetes version" }}
+{{- end }}
+{{- end }}
+
+{{- define "cluster.component.flatcar.version" }}
```
nprokopic, line 241: Similar like for `cluster.component.kubernetes.version`.

helm/cluster/templates/_helpers.tpl
```diff
@@ -151,4 +151,115 @@ Where `data` is the data to hash and `global` is the top level scope.
...
+{{- $flatcarVersion := include "cluster.component.version" $ | trimPrefix "v" }}
+{{- $flatcarVersion }}
+{{- else if $.GiantSwarm.providerIntegration.osImage }}
+{{- $.GiantSwarm.providerIntegration.osImage.version }}
+{{- else }}
+{{- fail "Cannot determine Flatcar version" }}
+{{- end }}
+{{- end }}
+
+{{- define "cluster.component.flatcar.variant" }}
```
nprokopic, line 254: Similar like for `cluster.component.flatcar.version`.

One caveat here, since `Release.spec.components[].version` must be a valid semver, and we use plain numbers for variants (like `1`, `2`, `3`), we would put something like `1.0.0`, `2.0.0`, `3.0.0`, and then just use major version number here.

Or alternatively, maybe better, we actually really use regular sember version number for variant number.

P.S. Variant (or the number in the image name suffix) is something that people just set manually when they build a custom image (by default the variant number is not set), so we can simply decide how it must look like when somebody does this.

helm/cluster/templates/apps/apps.yaml
```diff
@@ -70,7 +70,12 @@ spec:
       name: {{ $.Values.global.metadata.name }}-kubeconfig
       namespace: {{ $.Release.Namespace}}
   {{- end }}
-  version: {{ $ephemeralConfig.versionOverride | default $app.version | trim | trimPrefix "v" }}
+  {{- $appVersion := $app.version -}}
+  {{- if $.Values.providerIntegration.useReleases }}
+  {{- $_ := set $ "appName" $app.appName }}
+  {{ $appVersion = include "cluster.app.version" $ }}
+  {{- end }}
+  version: {{ $ephemeralConfig.versionOverride | default $appVersion | trim | trimPrefix "v" }}
```
nprokopic, line 78: Unrelated to Release CR, it is still possible to override app version like before (e.g. for the purposes of development, putting out on-call fire and similar).

helm/cluster/values.schema.json
```diff
@@ -1844,6 +1859,19 @@
...
                                 }
                             }
+                        },
+                        "offlineTesting": {
+                            "type": "object",
+                            "title": "Offline testing",
+                            "description": "Configure how Helm template rendering behaves when it is running during testing (e.g. locally or in the CI) and without connection to cluster.",
+                            "additionalProperties": false,
+                            "properties": {
+                                "renderWithoutReleaseResource": {
```
nprokopic, line 1869: A flag that enables rendering all manifests without Release CR existing. This is `false` by default, and it's set to true only in CI which is just rendering manifests without creating clusters.

njuettner (approved):

nprokopic:
Will merge PR after another round of testing.

PR #215, opened 2024-06-07
fix app chart names

bdehri:
### What does this PR do?

Fix `prometheus-blackbox-exporter` and `k8s-audit-metrics` apps to use the correct chart name


calvix:
are we consistent with the names across vintage and capi?

bdehri:
> are we consistent with the names across vintage and capi?

By taking a quick look at capa release CR and vintage release CR, I would say yes.

calvix (approved):

PR #228, opened 2024-06-13
Changes need to migrate from Chart Operator to Flux

ljakimczuk:
### What does this PR do?

In general it does two things:

* introduces new [Giantswarm Cluster Suite](https://github.com/giantswarm/giantswarm-cluster-suite) app that provides `giantswarm` Namespace and `giantswarm-critical` PriorityClass to the workload clusters. So far both of these resources were created by the Chart Operator.
* introduces flag to switch cluster from Chart Operator to Flux backend for App Platform. This flag gets translated into a label on Cluster CR which may then be processed by Cluster Apps Operator which is in charge of bootstrapping the platform.

Users will see the new app.

### How does it look like?

N/A

We are planning to open a way to migrate Chart Operator to Flux Helm Controller in App Platform.

### DIFF
--- a/helm/cluster/files/apps/k8s-audit-metrics.yaml
+++ b/helm/cluster/files/apps/k8s-audit-metrics.yaml
@@ -3,7 +3,8 @@ configKey: k8sAuditMetrics
 clusterValues:
   configMap: true
   secret: false
-dependsOn: kyverno
+dependsOnHelmRelease: true
+dependsOn: kyverno, cilium, prometheus-operator-crd
 namespace: kube-system
 # used by renovate
 # repo: giantswarm/k8s-audit-metrics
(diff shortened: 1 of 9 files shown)


helm/cluster/files/apps/k8s-audit-metrics.yaml
```diff
@@ -3,7 +3,8 @@ configKey: k8sAuditMetrics
 clusterValues:
   configMap: true
   secret: false
-dependsOn: kyverno
+dependsOnHelmRelease: true
+dependsOn: kyverno, cilium, prometheus-operator-crd
```
ljakimczuk, line 7: This app in fact depends on Cilium and Prometheus Operator CRD. With switch from Chart Operator to Flux it fails without these dependencies set.

PR #229, opened 2024-06-13
Fix app catalog and dependencies for new releases

nprokopic:
Towards https://github.com/giantswarm/roadmap/issues/3466

### Added

- Use app catalog from the Release CR if new releases are used
- Use app dependencies from the Release CR if new releases are used
- Add missing k8s-audit-metrics dependency (kyverno)

Clusters created with new releases will correctly use app dependencies from the Release CR.

### How does it look like?

No changes for customers.

https://github.com/giantswarm/roadmap/issues/3466

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,12 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Fixed
+
+- Use app catalog from the Release CR if new releases are used.
+- Use app dependencies from the Release CR if new releases are used.
+- Add missing k8s-audit-metrics dependency (kyverno).
+
 ## [0.32.0] - 2024-06-18
 
 ### Added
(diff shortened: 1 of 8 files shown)


nprokopic:
@Gacko the idea for `RELEASE_VERSION` var was to trigger usage of Release CR (that CAPA would use), or fallback to hard-coded versions in the chart (that all other providers would use, and our CI).

But I think another `USE_RELEASES` var is needed, so it's more clear. And then if `USE_RELEASES` is set, it looks for `RELEASE_VERSION` var, or it reads from the CI values.

nprokopic:
> the idea for `RELEASE_VERSION` var was to trigger usage of Release CR (that CAPA would use), or fallback to hard-coded versions in the chart (that all other providers would use, and our CI).

I was checking the rendered Helm templates for both cases, which is why I made this change in the Makefile, so we can more easily test both.

Gacko:
My suggestion was mostly about not duplicating code. This is why I simply introduced a new var which is getting used in any case, so you do not have two `helm template` commands which might divert over time.

nprokopic:
> My suggestion was mostly about not duplicating code. This is why I simply introduced a new var which is getting used in any case, so you do not have two `helm template` commands which might divert over time.

Let's improve the Makefile in a follow-up PR, as the changes here are needed for v25 release.

Gacko (approved):

njuettner (approved):

PR #244, opened 2024-07-02
enable statefulset pvc autodeletion feature gate on all clusters

QuantumEnigmaa:
### What does this PR do?

Towards https://github.com/giantswarm/giantswarm/issues/29575

This PR enables a feture gate that allows PVCs generated by statefulsets' pods to be automatically deleted when the corresponding pod is deleted. This allows to save on nodes' disk usage as well as preventing potential issues arising from leftover PVCs.

### How does it look like?


Gacko:
Eh, I think you pretty much destroyed the `values.schema.json` unintentionally. 🤣

Gacko:
Try implementing your changes to the default values in the default section of the `values.schema.json` and run `make generate` afterwards. That should do the trick. :)

Gacko (changes requested):
Try implementing your changes to the default values in the default section of the `values.schema.json` and run `make generate` afterwards. That should do the trick. 🙂

QuantumEnigmaa:
@Gacko It should be all good now :) 
But I must say that generating the values file from the schema is quite tiresome.

helm/cluster/values.schema.json
```diff
@@ -1735,7 +1735,30 @@
...
+                                                        "title": "Enabled",
+                                                        "description": "Enable the corresponding feature gate."
+                                                    },
+                                                    "name": {
+                                                        "type": "string",
+                                                        "title": "Name",
+                                                        "description": "Name of the feature gate."
+                                                    }
+                                                }
+                                            },
```
Gacko, line 1755:
```suggestion
                                            "$ref": "#/$defs/featureGates",
```

helm/cluster/values.schema.json
```diff
@@ -1745,7 +1768,30 @@
...
```
Gacko, line 1788:
```suggestion
                                            "$ref": "#/$defs/featureGates",
```

helm/cluster/whatever.schema.json
Gacko, line 1: I think you committed this accidentally, right? 🙂

helm/cluster/README.md
Gacko, line 1: Please re-generate the docs once you updated the schema. Thank you!

Gacko (changes requested):

helm/cluster/whatever.schema.json
QuantumEnigmaa: No it was intentional. I based myself on the other parts of the schema to write this.

QuantumEnigmaa: Oh no my bad this one yes :sweat_smile:

helm/cluster/values.schema.json
QuantumEnigmaa: Why shouldn't we have the full description ? According to what is done on the other parts of the schema, it shouldn't juste have a `$ref` right ?

Gacko: Because $ref is already referencing a schema definition with description and everything. This is where the docs are being generated from. Also that template is being used in other places. If you're not fine with the description of the definition, please update it there. But we'd like to use those definitions where possible to reduce code duplication.

QuantumEnigmaa: Ok got it. Replaced it as you suggested :)

QuantumEnigmaa:
@Gacko I think I addressed all your concerns :)

Gacko (approved):

PR #260, opened 2024-07-10
Use MC Zot as registry mirror by default in all workload clusters

nprokopic:
Towards https://github.com/giantswarm/roadmap/issues/3069

#### Changed

- Add MC Zot deployment as a registry mirror for `gsoci.azurecr.io` registries. This is the new default behavior for all workload clusters.

#### Added

- Tests for checking rendered containerd config.

In all new WCs, containerd will use MC Zot deployment and pull images from it by default.

### How does it look like?

Default registry mirrors configuration in containerd config now looks like this:

```toml
[plugins."io.containerd.grpc.v1.cri".registry]
  [plugins."io.containerd.grpc.v1.cri".registry.mirrors]
    [plugins."io.containerd.grpc.v1.cri".registry.mirrors."docker.io"]
      endpoint = ["https://registry-1.docker.io","https://giantswarm.azurecr.io",]
    [plugins."io.containerd.grpc.v1.cri".registry.mirrors."gsoci.azurecr.io"]
      endpoint = ["https://zot.giantmc.example.gigantic.io","https://gsoci.azurecr.io",]
```

- https://github.com/giantswarm/roadmap/issues/3069

Check if the implementation is correct. Please check newly added tests and if expected containerd configuration is correct there.


nprokopic:
Issue with Compare Helm rendering is known, it's because new Helm value is added, so it's treated as unknown when rendering templates on main branch (because new Helm values are incorrectly used with old templates).

helm/cluster/files/etc/containerd/config.toml
```diff
@@ -31,6 +31,9 @@ sandbox_image = "{{ include "cluster.image.registry" $ }}/{{ $.Values.providerIn
         {{- if and $.Values.global.components.containerd.localRegistryCache.enabled (has $host $.Values.global.components.containerd.localRegistryCache.mirroredRegistries) -}}
         "http://127.0.0.1:{{ $.Values.global.components.containerd.localRegistryCache.port }}",
         {{- end -}}
+        {{- if and $.Values.global.components.containerd.managementClusterRegistryCache.enabled (has $host $.Values.global.components.containerd.managementClusterRegistryCache.mirroredRegistries) -}}
+        "https://zot.{{ $.Values.global.managementCluster }}.{{ $.Values.global.connectivity.baseDomain }}",
+        {{- end -}}
         {{- range $value := $config -}}
         "https://{{$value.endpoint}}",
```
njuettner, line 38: just for my understanding, will this not add always to the last endpoint a ",". How can this work in the config?

Example:
CONFIG.TOML
...
zot.something.com,
myendpoint.something.com,

I assume the "," is only valid if another mirror registry is added, right?

helm/cluster/tests/components/test_containerd_config.go
```diff
@@ -0,0 +1,121 @@
+package components
```
njuettner, line 1: Maybe we could improve the code a bit by using testCases, because it's easier to read

```
package components

import (
    "bytes"
    "encoding/base64"
    "fmt"
    "os"
    "os/exec"
    "path/filepath"
    "runtime"
    "strings"

    . "github.com/onsi/ginkgo/v2"
    . "github.com/onsi/gomega"
)

// Embed test files
//go:embed testfiles/containerd_expected_default.toml
var expectedContainerdDefaultConfig string
//go:embed testfiles/containerd_expected_zot_mc.toml
var expectedContainerdZotMcConfig string
//go:embed testfiles/containerd_expected_zot_local.toml
var expectedContainerdZotLocalConfig string

var _ = Describe("containerd config", func() {
    type testCase struct {
        description    string
        helmValuesFile string
        expectedConfig string
    }

    // Define test cases
    testCases := []testCase{
        {
            description:    "when only required and default Helm values are used",
            helmValuesFile: "test-required-values.yaml",
            expectedConfig: expectedContainerdDefaultConfig,
        },
        {
            description:    "when MC zot is enabled",
            helmValuesFile: "test-zot-mc-values.yaml",
            expectedConfig: expectedContainerdZotMcConfig,
        },
        {
            description:    "when local zot is enabled",
            helmValuesFile: "test-zot-local-values.yaml",
            expectedConfig: expectedContainerdZotLocalConfig,
        },
    }

    for _, tc := range testCases {
        Context(tc.description, func() {
            It("renders the expected containerd config", func() {
                renderedConfig := renderContainerdConfig(tc.helmValuesFile)
                Expect(renderedConfig).To(Equal(strings.TrimSpace(tc.expectedConfig)), fmt.Sprintf("expected:\n>>>%s<<<\n, got:\n>>>%s<<<", tc.expectedConfig, renderedConfig))
            })
        })
    }
})

// render logic here similar to the JustBeforeEach block 
func renderContainerdConfig(helmValuesFile string) string {
...
}
```

uvegla:
We currently do not mirror `docker.io`with the MC zot, only `gsoci `on-demand: https://github.com/giantswarm/management-cluster-bases/blob/ff3f6771c0e27f734196be96af6a6cac308000e8/extras/zot-cache/values.yaml#L102-L117.

Also `gsoci` contains GS images, not mirroring / syncing `docker.io` in any ways to my knowledge.

helm/cluster/files/etc/containerd/config.toml
nprokopic: Trailing commas (ending an array with a comma, without an element behind it) are allowed in TOML, see here https://toml.io/en/v1.0.0#array.

> Arrays can span multiple lines. A terminating comma (also called a trailing comma) is permitted after the last value of the array.

helm/cluster/tests/components/test_containerd_config.go
nprokopic: Basically with Ginko tests are written in "Ginko style" with ginko and gomega concepts, while what we have been doing with test cases approach in some projects is more like a plain Golang approach to table specs, without following any test framework's design.

nprokopic:
> We currently do not mirror `docker.io`with the MC zot, only `gsoci `on-demand: https://github.com/giantswarm/management-cluster-bases/blob/ff3f6771c0e27f734196be96af6a6cac308000e8/extras/zot-cache/values.yaml#L102-L117.
> 
> Also `gsoci` contains GS images, not mirroring / syncing `docker.io` in any ways to my knowledge.

Thanks, will update this then.

nprokopic:
> > We currently do not mirror `docker.io`with the MC zot, only `gsoci `on-demand: https://github.com/giantswarm/management-cluster-bases/blob/ff3f6771c0e27f734196be96af6a6cac308000e8/extras/zot-cache/values.yaml#L102-L117.
> > Also `gsoci` contains GS images, not mirroring / syncing `docker.io` in any ways to my knowledge.
> 
> Thanks, will update this then.

Updated.

fiunchinho (approved):

CHANGELOG.md
```diff
@@ -7,6 +7,15 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Changed
+
+- Add MC Zot deployment as a registry mirror for `docker.io` and `gsoci.azurecr.io` registries. This is the new default behavior.
+- Add `gsoci.azurecr.io` as a mirror for `docker.io`.
```
uvegla, line 13: This needs to be updated.

helm/cluster/values.yaml
```diff
@@ -31,10 +31,16 @@ global:
         docker.io:
           - endpoint: registry-1.docker.io
           - endpoint: giantswarm.azurecr.io
```
uvegla, line 33: @piontec Should we get rid of `giantswarm.azurecr.io` on CAPx releases?

nprokopic: I left it here on purpose, because removing it would be a breaking change. I'm not sure if something somewhere relies on this. In any case, with a breaking change we would have to release this in a new major release, and now with v27 being already out, that would mean releasing this in v28 the soonest. And IIUC we want this in v25, in a new minor, so newly migrated AWS clusters get to use this. A new v25 minor cannot have a breaking change.

IMO we can remove this later, no harm in keeping the old defaults, while adding the new ones.

CHANGELOG.md
nprokopic: Updated 👍

uvegla: Do I get something wrong? `gsoci.azurecr.io` is not mirroring `docker.io`. It contains our images after a certain point (about 1 year ago). And the MC zot deployment is only a mirror for `gsoci.azurecr.io` at the moment.

Looking at your example here: https://github.com/giantswarm/cluster/pull/260#issue-2400412416 that looks good. But then this change log is not correct.

uvegla (approved):

CHANGELOG.md
nprokopic: I have updated the changelog, now it just has this:

```
### Changed

- Add MC Zot deployment as a registry mirror for `gsoci.azurecr.io` registry. This is the new default behavior.

### Added

- Tests for checking rendered containerd config.
```

paurosello:
how is it planned to inject the MC basedomain here? on kubectl-gs cluster creation?

nprokopic:
> how is it planned to inject the MC basedomain here? on kubectl-gs cluster creation?

Base domain is set on the MC level (for all WCs) in catalog values.

nprokopic:
Tests are green again :tada: https://github.com/giantswarm/cluster-aws/pull/711#issuecomment-2247307415

Merging this and releasing it!

PR #267, opened 2024-07-18
Feature Gates: Remove `CronJobTimeZone`.

Gacko:
This PR removes the `CronJobTimeZone` feature gate as it becomes stable and is included in Kubernetes v1.29.

For Kubernetes <v1.29, you will need to re-enable it using the respective values.

### DIFF
--- a/README.md
+++ b/README.md
@@ -130,8 +130,8 @@ cluster:
             apiAudiences:
               templateName: awsApiServerApiAudiences # name of the Helm template that renders api-audiences value
             featureGates:
-              - enabled: true
-                name: CronJobTimeZone
+            - name: StatefulSetAutoDeletePVC
+              enabled: true
             serviceAccountIssuer:
               clusterDomainPrefix: irsa # which sets service-account-issuer to irsa.<cluster name>.<base domain>
 ```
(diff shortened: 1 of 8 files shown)


README.md
```diff
@@ -130,8 +130,8 @@ cluster:
             apiAudiences:
               templateName: awsApiServerApiAudiences # name of the Helm template that renders api-audiences value
             featureGates:
-              - enabled: true
-                name: CronJobTimeZone
+            - name: StatefulSetAutoDeletePVC
+              enabled: true
```
njuettner, line 134: Why do we introduce a new feature gate here?
If it's intended I'd say we should mention it in the changelog.

Gacko: This is just in the `README.md` as an example of how customers can modify their cluster.

Gacko:
Note: This PR is going to be tested from the CAPA v29.0.0 PR: https://github.com/giantswarm/releases/pull/1318

I therefore put the test version of the chart into `cluster-aws` and put the test version of `cluster-aws` in the release as the normal cluster test suite cannot rely on a yet unreleased CAPA version.

njuettner (approved):

PR #272, opened 2024-07-23
Support pauseProperties with null values

glitchcrab:
### What does this PR do?

Adds support for pauseProperties with null values. This is useful for pausing the cluster until a value has been set.

The Cluster resource created for `cluster-vsphere` is correctly paused when it is initially deployed.

### How does it look like?

Ref: https://gigantic.slack.com/archives/C0559SH3RJ4/p1721749122430959


glitchcrab:
I have tested this using [this PR](https://github.com/giantswarm/cluster-vsphere/pull/243) and it functions exactly as expected. When the `host` value is null then the Cluster is paused, and when the `host` value contains an IP then the cluster is unpaused.

njuettner (approved):

nprokopic (approved):

PR #275, opened 2024-07-24
Use helper to render labels

fiunchinho:
### What does this PR do?

We already have a helper function to render all these labels, so let's just use it.

There should be no changes at all.

Check if final rendering is the same.

No

No


fiunchinho:
hey @giantswarm/team-turtles , would it make sense to add all `metadata.labels` to all CRs? CRs like the `MachinePool` don't get the `metadata.labels`, or the `KubeadmConfigTemplate`

Gacko (approved):

nprokopic:
> would it make sense to add all metadata.labels to all CRs?

Probably yes.

But maybe just check first if some of those would trigger rolling updates (e.g. for KubeadmConfigTemplate). AFAIK label changes should never trigger that, and e.g. KubeadmConfigTemplate label changes should be propagated in-place to existing KubeadmConfig CRs, but can't tell I'm 100% sure :shrug:

Gacko (approved):

PR #288, opened 2024-08-06
Add OS tooling named template

nprokopic:
### What does this PR do?

This PR adds `cluster.os.tooling.version` helper which is used to obtain the OS tooling (capi-image-builder) version that is then used to build the OS image name.

Also it adds few new and align names of existing OS-related named templates:
- `cluster.os.name` - new named template that returns the operating system name (fixed return value "flatcar").
- `cluster.os.releaseChannel` - new named template that returns the operating system release channel. It returns `.Values.providerIntegration.osImage.channel` which is "stable" by default for all providers (Helm value was already there, just the template is added here).
- `cluster.os.version` - just calls `cluster.component.flatcar.version` (while `cluster.component.flatcar.version` is marked as deprecated).

Giant Swarm engineers can add and work with `os-tooling` component in the Release CR.

This version will be always present and it will change whenever there is a new version of some binary that is baked into the OS image, as well as when capi-image-builder itself has some changes which requires building new images.

When writing manifests in cluster-\<provider\> charts, Giant Swarm engineers can now also use a set of public named templates from cluster chart (see above and see updated cluster chart docs). This way all changes in the implementation will happen in cluster chart only and will get applied to all providers by just updating cluster chart version.

### How does it look like?

In the Release CR:

```
  components:
  - name: cluster-aws
    catalog: cluster
    version: 1.3.0
  - name: flatcar
    version: 3815.2.5
  - name: os-tooling
    version: 1.14.4
```

https://github.com/giantswarm/roadmap/issues/3592

Yes :heavy_check_mark: 

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Added
 
 - Apps: Add `observability-policies` v0.0.1. ([#283](https://github.com/giantswarm/cluster/pull/283))
+- `cluster.os.name` named template that returns the operating system name.
+- `cluster.os.releaseChannel` named template that returns the operating system release channel.
+- `cluster.os.version` named template that returns the operating system version.
+- `cluster.os.tooling.version` named template which is used to obtain the OS tooling (capi-image-builder) version.
 
 ### Changed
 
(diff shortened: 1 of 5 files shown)


Gacko (approved):

nprokopic:
Testing cluster chart (and cluster-aws) changes here in the releases repo PR https://github.com/giantswarm/releases/pull/1337 (with a Release CR that is using cluster-aws dev version) :crossed_fingers:

Gacko (approved):

nprokopic:
Merging PR, as discussed in the Turtles standup.

PR #296, opened 2024-08-13
Add toggle to enable creation of giantswarm user on nodes

glitchcrab:
### What does this PR do?

Decouples creation of the `giantswarm` user on nodes from the bastion machinedeployment.

Nothing - the original chart behaviour is retained and the new behaviour is disabled by default (requires the provider chart integration to enable it on a per-cluster basis).

CAPV and CAPVCD may not use the bastion machinedeployment as the bastions are often managed outside of the provider charts, however we still need the `giantswarm` user to exist on the nodes.

### DIFF
--- a/helm/cluster/README.md
+++ b/helm/cluster/README.md
@@ -712,6 +712,7 @@ Provider-specific properties that can be set by cluster-$provider chart in order
 | `providerIntegration.environmentVariables.ipv4` | **IPv4** - Environment for IPv4.|**Type:** `string`<br/>|
 | `providerIntegration.hashSalt` | **Hash salt** - If specified, this token is used as a salt to the hash suffix of some resource names. Can be used to force-recreate some resources.|**Type:** `string`<br/>|
 | `providerIntegration.kubeadmConfig` | **Provider-specific kubeadm config** - Provider-specific kubeadm config that is common for all nodes, including both control plane and workers.|**Type:** `object`<br/>|
+| `providerIntegration.kubeadmConfig.enableGiantswarmUser` | **Enable giantswarm User** - If enabled, the giantswarm user will be created on all nodes.|**Type:** `boolean`<br/>**Default:** `false`|
 | `providerIntegration.kubeadmConfig.files` | **Files** - Provider-specific files that are deployed to all nodes. They are specified in the cluster-<provider> apps.|**Type:** `array`<br/>|
 | `providerIntegration.kubeadmConfig.files[*]` | **File from secret** - It defines a file with content in a Secret|**Type:** `object`<br/>|
 | `providerIntegration.kubeadmConfig.files[*].contentFrom` | **Content from** - It specifies where the file content is coming from.|**Type:** `object`<br/>|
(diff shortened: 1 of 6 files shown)


helm/cluster/README.md
```diff
@@ -781,6 +781,7 @@ Provider-specific properties that can be set by cluster-$provider chart in order
 | `providerIntegration.kubeadmConfig.postKubeadmCommands[*]` |**None**|**Type:** `string`<br/>|
 | `providerIntegration.kubeadmConfig.preKubeadmCommands` | **Pre-kubeadm commands** - Extra commands to run before kubeadm runs.|**Type:** `array`<br/>|
 | `providerIntegration.kubeadmConfig.preKubeadmCommands[*]` |**None**|**Type:** `string`<br/>|
+| `providerIntegration.kubeadmConfig.users` | **Users** - If enabled, the giantswarm user will be created on all nodes.|**Type:** `boolean`<br/>**Default:** `false`|
```
glitchcrab, line 784: the only thing i'm not sure about is the location of the newly-added toggle in the values - is this satisfactory for Team :turtle:?

helm/cluster/templates/clusterapi/_helpers_users.tpl
```diff
@@ -1,9 +1,7 @@
 {{- define "cluster.internal.kubeadm.users" }}
-{{- if $.Values.providerIntegration.resourcesApi.bastionResourceEnabled }}
-{{- if .Values.global.connectivity.bastion.enabled }}
+{{- if or (and $.Values.providerIntegration.resourcesApi.bastionResourceEnabled .Values.global.connectivity.bastion.enabled) $.Values.providerIntegration.kubeadmConfig.users }}
```
Gacko, line 2:
```suggestion
{{- if or (and .Values.providerIntegration.resourcesApi.bastionResourceEnabled .Values.global.connectivity.bastion.enabled) .Values.providerIntegration.kubeadmConfig.users }}
```

Gacko (approved):
Looks good to me, maybe align the `$` thingy. 🙂

Gacko (approved):

helm/cluster/values.schema.json
```diff
@@ -2567,6 +2567,12 @@
                         },
                         "preKubeadmCommands": {
                             "$ref": "#/$defs/preKubeadmCommands"
+                        },
+                        "users": {
```
nprokopic, line 2571: nitpick: I would probably expect that under `users` I would have to set a list of users.

And here for the flag approach maybe call this var `enableGiantswarmUser` or something like that.

Naming is hard 🙈

helm/cluster/README.md
nprokopic: I think that the location is fine, it's alongside other kubeadm things.

Other kubeadm properties here are I think mostly aligned with what is set in kubeadm (similar structures), so maybe check the other comment here for naming boolean vs list https://github.com/giantswarm/cluster/pull/296#discussion_r1716578448

helm/cluster/values.schema.json
glitchcrab: yeah I wasn't sure about the naming either tbh, I ended up with `users` because it aligns with that key in the ignition config. Changing it to `enableGiantswarmUser` would definitely make its usage more obvious.

nprokopic (approved):

PR #320, opened 2024-09-04
Allow configuring `node-cidr-mask-size` flag on `kube-controller-manager`

mnitchev:
### What does this PR do?

This allows setting the `node-cidr-mask-size` flag. The default is `/24` networks, however vintage default is `/25` which may create issues when migrating

None on existing users, default is the same.

### How does it look like?

There is an addition value now.

N/A

N/A

The docs in the readme have been updated, not sure how they relate to the giantswarm docs at the moment.

Yes

### DIFF
--- a/helm/cluster/values.schema.json
+++ b/helm/cluster/values.schema.json
@@ -1413,6 +1413,14 @@
                                             ],
                                             "maxItems": 1,
                                             "minItems": 1
+                                        },
+                                        "nodeCidrMaskSize": {
+                                            "type": "integer",
+                                            "title": "Node CIDR mask size",
+                                            "description": "The size of the mask that is used for the node CIDR. The node CIDR is a sub-range of the pod CIDR and so the mask size and pod CIDR must be chosen such that there is enough space for the maximum number of nodes in the cluster.",
+                                            "default": 24,
+                                            "maximum": 27,
+                                            "minimum": 16
                                         }
                                     }
                                 },
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Added
+
+- Allow configuring `kube-controller-manager` `--node-cidr-mask-size` flag.
+
 ## [1.3.0] - 2024-09-06
 
 ### Added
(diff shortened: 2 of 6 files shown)


helm/cluster/templates/clusterapi/controlplane/_helpers_initconfiguration.tpl
```diff
@@ -12,6 +12,7 @@ nodeRegistration:
     {{- end }}
     cloud-provider: external
     healthz-bind-address: 0.0.0.0
+    node-cidr-mask-size: {{ $.Values.global.connectivity.network.pods.nodeCidrMaskSize | default 24 }}
```
Gacko, line 15: No need to hard-code as long as nobody is setting the value to null elsewhere, right?

Gacko (changes requested):

helm/cluster/templates/clusterapi/controlplane/_helpers_initconfiguration.tpl
mnitchev: right, stupid copilot

mnitchev: fixed, I removed it the first time then changed the value from internal to this one and forgot to remove it from the new generated code :D

Gacko (approved):

helm/cluster/values.schema.json
```diff
@@ -1399,6 +1399,12 @@
                                             ],
                                             "maxItems": 1,
                                             "minItems": 1
+                                        },
+                                        "nodeCidrMaskSize": {
+                                            "type": "integer",
```
AndiDog, line 1404: We could ensure it's in some reasonable range. I invented those values, feel free to change to 1 and 32 🤷:

```suggestion
                                            "type": "integer",
                                            "minimum": 16,
                                            "maximum": 27,
```

AndiDog: Let's also maybe add a CI `ci/test-*-values.yaml` file which overrides these values, just to check this feature keeps working

mnitchev: I guess the theoretical usable number is /27 with 30 usable addresses, as we have around 17 pods in `kube-system` on control-plane nodes. It wouldn't be particularly _useful_, but definitely _usable_

mnitchev: In the other direction though, the maximum number of pods per node is 256. This is according to docs on a [deprecated flag though](https://kubernetes.io/docs/reference/command-line-tools-reference/kubelet/). The [KubeletConfiguration](https://kubernetes.io/docs/reference/config-api/kubelet-config.v1beta1/) parameters no longer specify a maximum. So considering that I think /16 is big enough as a maximum here :shrug:

paurosello (approved):

fiunchinho (approved):

Gacko (approved):

PR #323, opened 2024-09-05
Chart: Support multiple service account issuers.

AndiDog:
### What does this PR do?

For Vintage-to-CAPA migration purposes, we currently use the cluster-aws value `internal.migration.irsaAdditionalDomain` and unfortunately, customers store that in their GitOps. Not what was desired. Instead, allow setting multiple `--service-account-issuer` URLs as feature in the `cluster` chart (this PR) and then use it in the next CAPA v25.1.x release in order to migrate the next customer clusters without anything `internal.*`.

Since this is for the migration, this is based on cluster-v0.35.0, but I want to port this also to `main` afterwards. I created a branch `release-v0.35.x` from v0.35.0 in the hope that releasing this single commit as v0.35.1+ will work.

No `internal` values used by customers.

As developers, we get a more straightforward way of adding the API server command line arguments: instead of some hacky kubeadm config patching, we use a regular kubeadm patch file. (Later on, we'll even be able to use the array-based argument config introduced in kubeadm config v1beta4.)

This rolls control plane nodes because the patch file is added.

Towards https://github.com/giantswarm/giantswarm/issues/31589


Gacko (approved):

Gacko:
Sadly you'll need to back-port it to all minor releases we are using in the upgrade path from v25.0.0 to v29.0.0 by now. 😢

AndiDog:
Rebased onto v0.35.1

fiunchinho (approved):

PR #330, opened 2024-09-12
Allow configuring node-cidr-mask-size flag on kube-controller-manager

mnitchev:
### What does this PR do?

See https://github.com/giantswarm/cluster/pull/320

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Added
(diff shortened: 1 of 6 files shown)


mnitchev:
CI unfortunately still pushes this as `1.3.0-4091868870405b5d779206c4c90650dcb6731eaf`, so I need to work around this now :(

mnitchev:
Wait actually I don't think I need to do this. The release should just work this is just for the release

Gacko:
I changed the title because otherwise it ends up in the commit history like that and it wasn't adding a lot of information in the context of the release branch.

PR #343, opened 2024-09-23
Only use customNodeLabels if it's set

fiunchinho:
### What does this PR do?

I've seen a workload cluster that got created without setting `customNodeLabels`, and the kubelet arguments look like this (notice the trailing comma)
```
healthz-bind-address: 0.0.0.0
node-ip: ${COREOS_EC2_IPV4_LOCAL}
node-labels: ip=${COREOS_EC2_IPV4_LOCAL},role=worker,giantswarm.io/machine-pool=fer01-nodepool0,
v: "2"
```

With the change in this PR, the `customNodeLabels` are only appended to the argument if they are defined.

Users can now skip setting the `customNodeLabels` field, and there won't be a trailing comma in the kubelet arguments.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,6 +13,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Chart: Support multiple service account issuers.\
   Change `providerIntegration.controlPlane.kubeadmConfig.clusterConfiguration.apiServer.serviceAccountIssuer` to plural `providerIntegration.controlPlane.kubeadmConfig.clusterConfiguration.apiServer.serviceAccountIssuers` and render them in the specified order as `--service-account-issuer` parameters for the API server.
 
+### Changed
+
+- Only add the `customNodeLabels` value to the kubelet `node-labels` argument in the `KubeadmConfig` when `customNodeLabels` is defined.
+
 ## [1.3.0] - 2024-09-06
 
 ### Added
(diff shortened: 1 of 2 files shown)


fiunchinho:
I'm confused. The description for `customNodeLabels`  in the json schema says

> Deprecated: use nodeLabels instead

But I couldn't find that field :thinking_face: Also, the `customNodeLabels` field is not deprecated on `cluster-aws`.

AverageMarcus (approved):

CHANGELOG.md
```diff
@@ -13,6 +13,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Chart: Support multiple service account issuers.\
   Change `providerIntegration.controlPlane.kubeadmConfig.clusterConfiguration.apiServer.serviceAccountIssuer` to plural `providerIntegration.controlPlane.kubeadmConfig.clusterConfiguration.apiServer.serviceAccountIssuers` and render them in the specified order as `--service-account-issuer` parameters for the API server.
 
+### Fixed
```
Gacko, line 16:
```suggestion
### Changed
```

Gacko (approved):
Fine apart from the nit in the changelog. 🙂

AndiDog (approved):

fiunchinho:
[I wanted to clear this topic](https://github.com/giantswarm/cluster/pull/343#issuecomment-2367798319) before merging, just in case we need to fix the description of the field.

Gacko:
Well, this PR is the first change you see when `git blame`ing the schema: https://github.com/giantswarm/cluster/pull/70/files

It got actively renamed to `customNodeLabels` there. So I guess the deprecation note is wrong. Also there are no other occurrences of anything like "nodeLabels" in the whole repo.

Gacko:
`nodeLabels` was the old name. I assume Nikola wanted to keep `nodeLabels` and add a deprecation note there, but somehow decided different while working on the PR. Dunno, for now it is consistently named `customNodeLabels` everywhere.

fiunchinho:
I created https://github.com/giantswarm/cluster/pull/345

PR #348, opened 2024-09-25
Allow to configure cgroupsv1 per nodepool

fiunchinho:
### What does this PR do?

Towards https://github.com/giantswarm/roadmap/issues/3676

Users may configure different node pools with different cgroups configuration.
Also, control plane nodes always use cgroups v2.

Before this change, the configuration applied to all node pools and the control plane.

I could not find a good way of using the same `config.toml` template to render differently for the control plane nodes and the node pools. For some reason, Helm doesn't allow me to pass a custom `dict` of parameters when using the `tpl` function to render the `config.toml` file. I can't access subfields from the templat ewhen I do that, only fields at the root level. I tried different approaches, even asked in Slack, and we couldn't figure it out. So I ended up I following the principle of least astonishment having two different `config.toml` files: one for the control plane nodes, and another one for the worker nodes. It's not great but easy to understand. When customers finish migrating to cgroups v2, we don't need different configuration between node pools and control plane nodes, so at that moment we could remove this hack and keep only one config.toml file.


helm/cluster/templates/clusterapi/_helpers_files.tpl
```diff
@@ -37,7 +37,7 @@
 {{- end }}
 
 {{- define "cluster.internal.kubeadm.files.cgroupv1" }}
-{{- if $.Values.internal.advancedConfiguration.cgroupsv1 }}
+{{- if $.nodePool.config.cgroupsv1 }}
```
Gacko, line 40: Where is `$.nodePool` coming from here?

helm/cluster/values.schema.json
```diff
@@ -756,6 +756,12 @@
                         }
                     }
                 },
+                "cgroupsv1": {
+                    "type": "boolean",
+                    "title": "Cgroups v1",
```
Gacko, line 761: Please don't tell me it's a capital C because some validation requires this... 🙄🤣

helm/cluster/values.schema.json
```diff
@@ -1693,7 +1696,7 @@
                         "cgroupsv1": {
                             "type": "boolean",
                             "title": "CGroups v1",
-                            "description": "Force use of CGroups v1 for whole cluster.",
+                            "description": "Force use of CGroups v1 for whole cluster. Deprecated: use the node pool setting instead.",
```
Gacko, line 1699:
```suggestion
                            "description": "Force use of CGroups v1 for whole cluster. Deprecated: Use the node pool level setting instead.",
```

helm/cluster/values.schema.json
Gacko: Also: Would it make sense, to have this as a string? Just in case something like a v3 is coming one day.

CHANGELOG.md
```diff
@@ -7,6 +7,11 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Changed
+
+- Allow to configure `cgroups` v1 or v2 compatibility per node pool, instead of the whole cluster.
+- Control plane nodes always use cgroups v2.
```
Gacko, line 13: Why a new line here? Wouldn't it make more sense to have it in the previous line?

helm/cluster/templates/clusterapi/_helpers_files.tpl
AndiDog:
```
{{- $_ := set $ "nodePool" (dict "name" $nodePoolName "config" $nodePoolConfig) }}
```

helm/cluster/values.schema.json
fiunchinho: It's not like we just put the value of the variable in the configuration, but depending on the boolean value we run some logic to render different things. So I don't think we get much value out of making it a string, If a v3 comes along, we don't know what changes that may require.

helm/cluster/templates/clusterapi/_helpers_files.tpl
fiunchinho: If you start to follow the actual chain of function calls, templates rendering and so on, you could start forgetting who you are, but also [find it here](https://github.com/giantswarm/cluster/blob/main/helm/cluster/templates/clusterapi/workers/kubeadmconfig.yaml#L16).

helm/cluster/values.schema.json
Gacko: Sure, sure, I was just thinking of some kind of "switch case" logic on the string then. That would make it more flexible than having to change all the code instead of just adding another `case`.

fiunchinho:
[CAPA e2e tests passing](https://github.com/giantswarm/cluster-aws/pull/866#issuecomment-2382593404)

helm/cluster/files/etc/containerd/controlplane-config.toml
```diff
@@ -19,7 +19,7 @@ no_shim = false
 # setting runc.options unsets parent settings
 runtime_type = "io.containerd.runc.v2"
 [plugins."io.containerd.grpc.v1.cri".containerd.runtimes.runc.options]
-SystemdCgroup = {{ if $.Values.internal.advancedConfiguration.cgroupsv1 }}false{{else}}true{{end}}
+SystemdCgroup = true
```
fiunchinho, line 22: Always use cgroups v2 in the control plane.

helm/cluster/templates/clusterapi/_helpers_files.tpl
```diff
@@ -2,9 +2,7 @@
 {{- include "cluster.internal.kubeadm.files.sysctl" $ }}
 {{- include "cluster.internal.kubeadm.files.selinux" $ }}
 {{- include "cluster.internal.kubeadm.files.systemd" $ }}
-{{- include "cluster.internal.kubeadm.files.cgroupv1" $ }}
```
fiunchinho, line 5: All the removed lines in this file were moved to the workers helpers file , because now it's only used for worker nodes.

helm/cluster/README.md
```diff
@@ -289,7 +289,7 @@ For Giant Swarm internal use only, not stable, or not supported by UIs.
 | `internal.advancedConfiguration` | **Advanced configuration** - Advanced configuration of cluster components, to be configured by Giant Swarm staff only.|**Type:** `object`<br/>|
 | `internal.advancedConfiguration.appPlatform` | **App Platform** - Advanced configuration of App Platform.|**Type:** `object`<br/>|
 | `internal.advancedConfiguration.appPlatform.fluxBackend` | **Flux Backend** - Use Flux as App Platform backend for installing apps.|**Type:** `boolean`<br/>**Default:** `false`|
-| `internal.advancedConfiguration.cgroupsv1` | **CGroups v1** - Force use of CGroups v1 for whole cluster.|**Type:** `boolean`<br/>**Default:** `false`|
+| `internal.advancedConfiguration.cgroupsv1` | **CGroups v1** - Force use of CGroups v1 for whole cluster. Deprecated: use the node pool setting instead.|**Type:** `boolean`<br/>**Default:** `false`|
```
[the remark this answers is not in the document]
fiunchinho: Right! Thanks

helm/cluster/templates/clusterapi/workers/_helpers_files.tpl
```diff
@@ -1,5 +1,7 @@
 {{- define "cluster.internal.workers.kubeadm.files" }}
 {{- include "cluster.internal.kubeadm.files" $ }}
+{{- include "cluster.internal.kubeadm.files.cgroupv1" $ }}
```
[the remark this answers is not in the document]
fiunchinho: Good catch! thanks.

helm/cluster/files/etc/containerd/workers-config.toml
```diff
@@ -0,0 +1,56 @@
...
+runtime = "runc"
+# do not use a shim when starting containers, saves on memory but
+# live restore is not supported
+no_shim = false
+
+[plugins."io.containerd.grpc.v1.cri".containerd.runtimes.runc]
+# setting runc.options unsets parent settings
+runtime_type = "io.containerd.runc.v2"
+[plugins."io.containerd.grpc.v1.cri".containerd.runtimes.runc.options]
+SystemdCgroup = {{ if $.nodePool.config.cgroupsv1 }}false{{else}}true{{end}}
```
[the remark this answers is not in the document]
fiunchinho: Good catch! I wanted to use while the field is deprecated. We can removed when we remove the field.

fiunchinho:
I'm currently blocked because e2e tests won't use the right version https://gigantic.slack.com/archives/C07KSM2E51A/p1727967655453349

AndiDog (approved):

Gacko:
Tests in `cluster-aws` are working again and are actually green. Feel free to merge!

AndiDog (approved):

fiunchinho:
@Gacko I need approval from a code owner. Could you approve?

Gacko (approved):
Oh, sorry! I forgot about this... so sorry! Here's your approval, I'm really happy to see this merged! 🥳

PR #351, opened 2024-09-27
Enable templating within systemd unit files

AverageMarcus:
### What does this PR do?

Fixes https://github.com/giantswarm/roadmap/issues/3691

- Added support for `Service` section in custom systemd configuration
- Added support for `After`, `Requires`, `Wants` and `BindsTo` within unit section of custom systemd configuration
- Added support for passing Helm templating from provider chart values through to systemd unit templates

N/A - allows for additional configuration from cluster charts

### How does it look like?

N/A

Required by `cluster-cloud-director` to allow the customer to provide some custom values that will get added to the ignition


helm/cluster/templates/clusterapi/_helpers_flatcar.tpl
```diff
@@ -21,6 +21,23 @@
     {{- if hasKey .contents.unit "defaultDependencies" }}
     DefaultDependencies={{ if .contents.unit.defaultDependencies }}yes{{ else }}no{{ end }}
     {{- end }}
+    {{- if hasKey .contents.unit "after" }}
```
Gacko, line 24: I think `hasKey` has been used for the boolean value only. Here you should be fine with...

```suggestion
    {{- if hasKey .contents.unit.after }}
```

Also: wasn't it possible to have more than one `After` in a unit file? I'm not sure... so maybe it would be worth having an array here.

The same of course applies to the other fields.

AverageMarcus: It shouldn't make a difference. `haskey` works on any type and will return true if the object has anything with that key. Though thinking about it we likely want to avoid empty strings so I'll switch it out.

AverageMarcus: Updated

Gacko: Yeah, my concern was about empty strings. Sorry for the confusion!

njuettner (approved):
LGTM if tested

glitchcrab (approved):
i have no issues with this, so me likey

glitchcrab:
following up to my review, i thought of a helpful addition. as we're making an interface to systemd units here, it would be helpful to have `Wants` as an option in the `[Unit]` section please :pray:

glitchcrab:
And a further thought - would it be possible to support multiple Unit ordering fields? e.g. we have some units which have multiple `After` dependencies

Gacko (approved):

glitchcrab:
this was tested with the following provider values:

```
global:
  connectivity:
    network:
      staticRoutes:
        - destination: 10.2.3.0/24
          via: 10.9.8.7
        - destination: 10.20.30.0/24
          via: 10.9.8.7
```

and the following systemd unit values:

```
                - contents:
                    install:
                      wantedBy:
                        - multi-user.target
                    service:
                      additionalFields: |-
                        ExecStart=/usr/bin/bash -cv 'echo "$("$(find /usr/bin /usr/share/oem -name vmtoolsd -type f -executable 2>/dev/null | head -n 1)" --cmd "info-get guestinfo.ignition.network")" > /opt/bin/set-networkd-units.sh'
                        ExecStart=/usr/bin/bash -cv 'chmod u+x /opt/bin/set-networkd-units.sh'
                        ExecStart=/opt/bin/set-networkd-units.sh
                        {{- if $.global.connectivity.network.staticRoutes }}
                        ExecStart=/usr/bin/bash -cv 'sleep 3'
                        {{- range $.global.connectivity.network.staticRoutes }}
                        ExecStart=/usr/bin/bash -cv 'ip route add {{ .destination }} via {{ .via }}'
                        {{- end }}
                        {{- end }}
                      remainAfterExit: "yes"
                      type: oneshot
                    unit:
                      after:
                        - set-hostname.service
                      description: Install the systemd-networkd unit files and add static routes
                      requires:
                        - coreos-metadata.service
```

which produced the following unit:

```
                - name: set-networkd-units.service
                  enabled: true
                  contents: |
                    [Unit]
                    Description=Install the systemd-networkd unit files and add static routes
                    After=set-hostname.service
                    Requires=coreos-metadata.service
                    [Service]
                    Type=oneshot
                    RemainAfterExit=yes

                    ExecStart=/usr/bin/bash -cv 'echo "$("$(find /usr/bin /usr/share/oem -name vmtoolsd -type f -executable 2>/dev/null | head -n 1)" --cmd "info-get guestinfo.ignition.network")" > /opt/bin/set-networkd-units.sh'
                    ExecStart=/usr/bin/bash -cv 'chmod u+x /opt/bin/set-networkd-units.sh'
                    ExecStart=/opt/bin/set-networkd-units.sh
                    ExecStart=/usr/bin/bash -cv 'sleep 3'
                    ExecStart=/usr/bin/bash -cv 'ip route add 10.2.3.0/24 via 10.9.8.7'
                    ExecStart=/usr/bin/bash -cv 'ip route add 10.20.30.0/24 via 10.9.8.7'
                    [Install]
                    WantedBy=multi-user.target
```

AverageMarcus (commented):
This is beautiful! 😍  (I can't approve it though as I opened the PR 🤦 )

@Gacko Mind giving this work of art a little ✅  ?

anvddriesch (approved):

Gacko (approved):
LGTM!

PR #355, opened 2024-10-01
Added support for Before and PartOf within unit section of custom systemd configuration

glitchcrab:
### What does this PR do?

- Further extends the fields available for templating systemd units.
- Switch systemd Service ExecStart to an array to support multiple commands (added in #351)

None, backwards compatible addition which is configured by provider charts.

### How does it look like?

N/A

See https://github.com/giantswarm/cluster/pull/351


Gacko (approved):

helm/cluster/templates/clusterapi/_helpers_flatcar.tpl
```diff
@@ -42,10 +52,18 @@
     {{- end }}
     {{- if .contents.service }}
     [Service]
+    {{- if .contents.service.environment }}
```
Gacko, line 55: You could also repeat this, right? Like, `environment` could be an array?

glitchcrab:
ok I'm done adding stuff, this is ready to be reviewed please :pray:

helm/cluster/templates/clusterapi/_helpers_flatcar.tpl
glitchcrab: yes that's a good point - i'll switch that over

glitchcrab:
I decided that I might as well add support for `EnvironmentFile` while I'm at it too.

AverageMarcus (approved):

vxav (approved):

anvddriesch (approved):

PR #367, opened 2024-10-10
Allow to configure cgroupsv1 per nodepool

fiunchinho:
### What does this PR do?

Backport of https://github.com/giantswarm/cluster/pull/348

### DIFF
--- a/helm/cluster/templates/containerd.yaml
+++ b/helm/cluster/templates/containerd.yaml
@@ -1,8 +1,20 @@
-{{- if or $.Values.providerIntegration.resourcesApi.controlPlaneResourceEnabled $.Values.providerIntegration.resourcesApi.machinePoolResourcesEnabled }}
+{{- if $.Values.providerIntegration.resourcesApi.machinePoolResourcesEnabled }}
+{{- range $nodePoolName, $nodePoolConfig := $.Values.global.nodePools | default $.Values.providerIntegration.workers.defaultNodePools }}
+{{- $_ := set $ "nodePool" (dict "name" $nodePoolName "config" $nodePoolConfig) }}
 apiVersion: v1
 kind: Secret
 metadata:
-  name: {{ include "cluster.resource.name" $ }}-containerd-{{ include "cluster.data.hash" (dict "data" (tpl ($.Files.Get "files/etc/containerd/config.toml") $) "salt" $.Values.providerIntegration.hashSalt) }}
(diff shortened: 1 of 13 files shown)


glitchcrab:
please ignore me, wrong browser tab :man_facepalming:

Gacko:
I changed the PR's title as there is, IMHO, no need to prefix it this way. Apart from that: LGTM! I guess you only cherry-picked, right?

I guess it's currently hard to impossible to test stuff like this by just running cluster test suites, so we might need to come up with a test release that uses the test version of `cluster-aws` that includes this version of `cluster` here.

helm/cluster/templates/containerd.yaml
```diff
@@ -1,8 +1,20 @@
-{{- if or $.Values.providerIntegration.resourcesApi.controlPlaneResourceEnabled $.Values.providerIntegration.resourcesApi.machinePoolResourcesEnabled }}
+{{- if $.Values.providerIntegration.resourcesApi.machinePoolResourcesEnabled }}
+{{- range $nodePoolName, $nodePoolConfig := $.Values.global.nodePools | default $.Values.providerIntegration.workers.defaultNodePools }}
+{{- $_ := set $ "nodePool" (dict "name" $nodePoolName "config" $nodePoolConfig) }}
 apiVersion: v1
 kind: Secret
 metadata:
-  name: {{ include "cluster.resource.name" $ }}-containerd-{{ include "cluster.data.hash" (dict "data" (tpl ($.Files.Get "files/etc/containerd/config.toml") $) "salt" $.Values.providerIntegration.hashSalt) }}
+  name: {{ include "cluster.resource.name" $ }}-{{ $nodePoolName }}-containerd-{{ include "cluster.data.hash" (dict "data" (tpl ($.Files.Get "files/etc/containerd/workers-config.toml") $) "salt" $.Values.providerIntegration.hashSalt) }}
```
AndiDog, line 7: Let's hope nobody calls their node pool `controlplane` 😆

AndiDog (approved):

helm/cluster/templates/containerd.yaml
fiunchinho: lol I didn't think of that :smile: I don't think it will happen though (infamous last words)

fiunchinho:
I migrated a test cluster that was using cgroupsv1 in one of the node pools, and the configuration was migrated successfully.

PR #368, opened 2024-10-10
Allow to configure cgroupsv1 per nodepool (#348)

fiunchinho:
* Allow to configure cgroupsv1 per nodepool

* Apply suggestions from code review

* Regenerate docs

* Move flatcar cgroupsv1 function to worker helper

* Adjust tests to new containerd cgroups configuration

* Apply suggestions from code review

* Still support old field

---------

### What does this PR do?

### How does it look like?

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Changed
+
+- Allow configuring `cgroups` v1 or v2 compatibility per node pool, instead of the whole cluster. Control plane nodes always use cgroups v2.
+
 ## [1.0.2] - 2024-09-30
 
 ### Added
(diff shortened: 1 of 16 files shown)


Gacko:
I changed the PR's title as there is, IMHO, no need to prefix it this way. Apart from that: LGTM! I guess you only cherry-picked, right?

I guess it's currently hard to impossible to test stuff like this by just running cluster test suites, so we might need to come up with a test release that uses the test version of `cluster-aws` that includes this version of `cluster` here.

AndiDog (approved):

fiunchinho:
> I changed the PR's title as there is, IMHO, no need to prefix it this way. Apart from that: LGTM! I guess you only cherry-picked, right?
> 
> I guess it's currently hard to impossible to test stuff like this by just running cluster test suites, so we might need to come up with a test release that uses the test version of `cluster-aws` that includes this version of `cluster` here.

Yeah, basically cherry-picked and fixed conflicts.

Gacko (approved):

PR #385, opened 2024-11-13
Teleport: Add initial join token & handling for it on service level

ssyno:
### What does this PR do?
Enables SSH access to nodes before Kubernetes is ready by introducing an initial Teleport join token. Once Kubernetes is ready, the service restarts to use the token from secret.

### Effect on users
- SSH access available immediately during node bootstrap
- No changes needed for existing clusters
- Optional `initialJoinToken` for new clusters needing early access

### Background
Currently, nodes require Kubernetes to be ready and nodes rolling before allowing SSH access. 
This creates issues when troubleshooting Kubernetes nodes while bootstraping.

### Reviewer focus needed on
- Token transition reliability
- Service restart handling

### Release notes
```markdown
- Added support for early SSH access during node bootstrap
```


Gacko (commented):
Can you please elaborate on what's the current issue and how exactly this change improves the situation? I'm not a lot into Teleport and how it's being setup and so it would probably be interesting for everyone reading this PR to get some context.

ssyno:
> Can you please elaborate on what's the current issue and how exactly this change improves the situation? I'm not a lot into Teleport and how it's being setup and so it would probably be interesting for everyone reading this PR to get some context.

sure I going to update the PR

Gacko:
I somehow feel like the other PR obsoletes this one. Or at least you created the other one because this approach didn't work? Anyway: The other PR is clear to me. This one here is not, yet.

PR #387, opened 2024-11-14
Teleport: Add initial join token.

ssyno:
### What does this PR do?
Enables early SSH access by creating the Teleport token file during node boot. Kubernetes automatically takes over the token file by mounting the secret over it when ready.

### Effect on users
- SSH access available immediately during node bootstrap
- No service restarts needed
- Clean transition to secret-based tokens
- No changes needed for existing clusters

### Background
Currently, nodes require Kubernetes to be ready and nodes rolling before allowing SSH access.
This creates issues when troubleshooting Kubernetes nodes while bootstraping.

### Reviewer focus needed on
- Secret mount behavior

### Release notes
```markdown
- Added support for early SSH access during node bootstrap
```


Gacko:
Ok, now the confusion is complete. What's the point of having two different PRs, which - to me as someone who's not very familiar with the Teleport agent - look the same or at least feel like they could be one? 😅

Gacko (approved):
Anyway. This change on its own looks good to me! :) Let's discuss in the other PR.

Gacko (approved):

Gacko (approved):

ssyno:
Ready to be reviewed

helm/cluster/templates/clusterapi/_helpers_flatcar.tpl
```diff
@@ -263,9 +281,19 @@
     [Unit]
     Description=Teleport Service
     After=network.target
+    {{- if $.Values.providerIntegration.teleport.initialJoinToken }}
+    After=network.target teleport-init.service
+    Requires=teleport-init.service
+    RequiresMountsFor=/etc/teleport.yaml /etc/teleport-join-token /opt/teleport-node-role.sh
+    {{- end }}
```
Gacko, line 288: I _think_ it is valid to mention multiple `After`, but do you think it would be useful to combine them?

```suggestion
    {{- if $.Values.providerIntegration.teleport.initialJoinToken }}
    After=network.target teleport-init.service
    Requires=teleport-init.service
    RequiresMountsFor=/etc/teleport.yaml /etc/teleport-join-token /opt/teleport-node-role.sh
    {{- else }}
    After=network.target
    {{- end }}
```

ssyno: Totally agree with you, it will make it clearer.
committing suggestion

ssyno:
> LGTM if tested!

Tested in `cluster-aws` using `mc-bootstrap` if you want you can take a look on my PRs there too

PR #410, opened 2025-01-09
Set install and upgrade retries of HelmReleases to -1

ubergesundheit:
### What does this PR do?

- Set `spec.install.remediation.retries` to `-1` from previous `30`
- Set `spec.upgrade.remediation.retries` to `-1`. This wasn't set at all previously

This is a follow up to PR https://github.com/giantswarm/cluster/pull/394 for https://github.com/giantswarm/roadmap/issues/3664


weatherhog (approved):

Gacko (approved):

Gacko:
Thank you! No need to release this, yet, as it would make it way to clusters when released in the provider specific cluster app and a platform release, only.

PR #454, opened 2025-03-04
Apps: Remove source config from external-dns defaults.

mcharriere:
Signed-off-by: Matias Charriere <matias@giantswarm.io>

### What does this PR do?

remove source config from external-dns values

Default value changed in the latest app, but this is overriding.


mcharriere:
> So this change will probably end up in a v30.1.0 earliest.

That's fine, we can still override any cluster we need this configured on. Thank you!

PR #476, opened 2025-03-27
Enable different KSM custom resources in observability-bundle

iuriaranda:
### What does this PR do?

It will enable different custom resources in KSM depending on different conditions:

- if there's karpenter it will enable the karpenter CRs (note that this is not yet ready. Will be part of cluster-aws with https://github.com/giantswarm/cluster-aws/pull/1038)
- if it's an MC it will enable the different CAPI provider CRs

Towards https://github.com/giantswarm/giantswarm/issues/32502

### How does it look like?

### DIFF
--- a/helm/cluster/files/apps/observability-bundle.yaml
+++ b/helm/cluster/files/apps/observability-bundle.yaml
@@ -14,7 +14,7 @@ version: 1.12.0-632d76d83cf2f25a07bd8aa933891ea5848fc384
 catalog: default-test
 defaultValues:
   kubeStateMetricsCustomResources:
-    {{- if include "cluster.internal.isManagementCluster" $ }}
+    {{- if eq (include "cluster.internal.isManagementCluster" $) "true" }}
     flux_kustomization_v1: true
     flux_helmrelease_v2beta1: true
     flux_gitrepository_v1: true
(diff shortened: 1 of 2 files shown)


helm/cluster/files/apps/observability-bundle.yaml
```diff
@@ -11,3 +11,44 @@ namespace: kube-system
 # used by renovate
 # repo: giantswarm/observability-bundle
 version: 1.12.0
+defaultValues:
+  kubeStateMetricsCustomResources:
+    {{- if include "cluster.internal.hasKarpenterNodePool" $ }}
+    karpenter_machine_v1: true
```
QuentinBisson, line 17: This should be aws only right?

helm/cluster/templates/_helpers.tpl
```diff
@@ -522,3 +522,28 @@ Where `data` is the data to hash and `global` is the top level scope.
...
+    {{- break -}}
+  {{- end -}}
+{{- end -}}
+{{- end -}}
+
+{{/*
+  cluster.internal.isMC is a public named template that returns true if the cluster is a management cluster, otherwise
+  it returns false.
+*/}}
+{{- define "cluster.internal.isMC" -}}
```
QuentinBisson, line 543:
```suggestion
{{- define "cluster.internal.isManagementCluster" -}}
```

helm/cluster/files/apps/observability-bundle.yaml
iuriaranda: For now yes, but potentially karpenter could be deployed on other providers like Azure. This check is useless atm, until [this PR](https://github.com/giantswarm/cluster-aws/pull/1038) is merged, since we don't deploy karpenter as a cluster app yet. So I think I'll remove it for now and add it as part of the karpenter integration effort later on

QuentinBisson: You can keep it it's fine to me as long as it does not screw up ksm :)

QuentinBisson:
Can you endure customers can still add their own custom resource définition ?

helm/cluster/files/apps/observability-bundle.yaml
iuriaranda: well since the karpenter integration is not yet final, I'm not 100% sure the check will work as it is anyway. So I prefer to just remove it for now

PR #478, opened 2025-04-01
Support for region-specific container registry mirror configuration via `mirrorsTemplateName` property

njuettner:
### What does this PR do?

Towards: https://github.com/giantswarm/roadmap/issues/3887

Added `mirrorsTemplateName` property to configure region-specific registry mirrors which can be used by provider charts to change default registry mirror configuration, e.g. China.

This would be integrated by setting a provider specific helm helper function, e.g.:
```yaml
{{- define "awsContainerMirrorRegistry" -}}
{{- if hasPrefix "cn-" (include "aws-region" $ ) -}}
docker.io:
  - endpoint: foo.something.com
  - endpoint: bar.something.com
another.registry.io:
  - endpoint: foo.something.com
  - endpoint: bar.something.com
{{- else -}}
{{- $.Values.global.components.containerd.containerRegistries | toYaml -}}
{{- end -}}
{{- end -}}
```

### Testing

**Also a big fat note on the test**, I currently skipped the rendering of Helm template for `helm/cluster/ci/test-registry-mirrors-values.yaml` on old code which we can change after the merge again. 

When running `helm template -n org-giantswarm -f "old/helm/cluster/ci/ci-values.yaml" -f "${test_file_path}" "old/helm/cluster"` 

It's trying to render the chart from the old/helm/cluster directory, but the custom template test helpers only exist in the current working directory, not in the old code.


helm/cluster/templates/_test_registry_helpers.tpl
```diff
@@ -0,0 +1,21 @@
+{{/* Test helpers for registry configuration */}}
+{{- define "cluster.internal.test.awsContainerImageRegistry" -}}
```
AverageMarcus, line 2: Where is this template used? I'm not seeing it in here.

AverageMarcus (approved):
LGTM but I can't see where one of the templates is used

helm/cluster/templates/_test_registry_helpers.tpl
njuettner: It's defined in helm/cluster/ci/test-registry-mirrors-values.yaml

AverageMarcus: Oh this is test specific? 🤦

njuettner: Yes 👍🏻

fiunchinho (approved):
cool!

PR #505, opened 2025-05-13
Enable setting taints on (all/worker/controller) nodes through provider integration

robinkb:
### What does this PR do?

Enables setting taints on worker nodes, controller nodes, or all of them, by adding the following values:

* `providerIntegration.kubeadmConfig.taints`
* `providerIntegration.controlPlane.kubeadmConfig.taints`
* `providerIntegration.workers.kubeadmConfig.taints`

These are fed to the `KubeadmConfiguration` for all nodes created through the chart.

Users are able to set default taints on nodes.

### How does it look like?

Success.

Will enable Phoenix to solve https://github.com/giantswarm/giantswarm/issues/31687.

We discussed this in Slack, so it should be fine. :crossed_fingers: 

Yup, updated the generated docs.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 - Support for region-specific container registry mirror configuration via `mirrorsTemplateName` property.
 - Make `ntpd` configurable.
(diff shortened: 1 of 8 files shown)


AverageMarcus (approved):
Love it!

PR #508, opened 2025-05-14
Configure the kubelet to register CP nodes with the node-role.kubernetes.io/control-plane taint

iuriaranda:
### What does this PR do?

This is necessary during disaster recovery scenarios, where we restore an etcd snapshot on a running cluster and we need to restart the kubelet to register an existing CP node to the recovered cluster state. Without this the CP node would get registered without the taint, allowing any Pod to get scheduled on it.

For context, normally this taint is added automatically by `kubeadm init` or `kubeadm join`, which will just proceed normally if it's already added by the kubelet before. Kubeadm is not involved during a DR scenario, since it's not designed to be re-run on an already bootstrapped node.

The CP nodes will roll

### How does it look like?

Related to https://github.com/giantswarm/giantswarm/issues/30439

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,6 +15,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   - `providerIntegration.kubeadmConfig.taints`
   - `providerIntegration.controlPlane.kubeadmConfig.taints`
   - `providerIntegration.workers.kubeadmConfig.taints`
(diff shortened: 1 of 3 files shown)


njuettner (approved):

fiunchinho:
I just saw this PR. I have one question, is there any difference or benefit on using this approach vs using `Values.global.controlPlane.customNodeTaints`?

PR #553, opened 2025-07-17
Configure cilium to use the k8s hostname defined in cluster-info CM

iuriaranda:
### What does this PR do?

Right now, cilium uses the external public hostname of the cluster to access the API server. This makes a critical internal component dependant on the baseDomain DNS infrastructure, which is sometimes managed by the customer, sometimes managed by us, and could be unreliable. Additionally it makes baseDomain more rigid and difficult to change on an existing cluster.

When setting `k8sServiceHost: auto`, the cilium helm chart will fetch the k8s API hostname from the `cluster-info` ConfigMap in the `kube-public` Namespace, which is set to an internal hostname or IP inside the provider network (e.g. the ELB hostname in AWS and Azure).

None

### How does it look like?

https://gigantic.slack.com/archives/CLPMFRVU6/p1752252511295269

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Changed
 
 - Set defaults for net-exporter App.
+- Configure cilium to use the k8s hostname defined in the cluster-info ConfigMap
 
 ## [2.4.0] - 2025-06-03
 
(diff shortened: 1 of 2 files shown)


Gacko (approved):

Gacko:
LGTM, but please let @giantswarm/team-cabbage have a second look. 🙂

Also I can relate to and appreciate this change, because I recently fell for this, as in discovered Cilium needs to know the API server LB FQDN as it provides the `kube-proxy` replacement and therefore the default `kubernetes` service doesn't work for it. 😄

iuriaranda:
Tests are passing on all providers, so we are good to go!

PR #570, opened 2025-08-11
allow to disable worker kubeadmconfig

anvddriesch:
### What does this PR do?
this adds a field that allows to disable rendering the kubeadmconfig cr if its not needed (eg eks)
https://github.com/giantswarm/giantswarm/issues/33993

### How does it look like?

### DIFF
--- a/.github/workflows/cluster-provider-test-pull-request.yaml
+++ b/.github/workflows/cluster-provider-test-pull-request.yaml
@@ -11,7 +11,7 @@ jobs:
 
     steps:
     - name: Checkout repository
-      uses: actions/checkout@v4
+      uses: actions/checkout@v5
 
     - name: Set up Git
       run: |
--- a/.github/workflows/test-manifests.yaml
+++ b/.github/workflows/test-manifests.yaml
@@ -13,7 +13,7 @@ jobs:
 
     steps:
     - name: Checkout repository
-      uses: actions/checkout@v4
+      uses: actions/checkout@v5
       with:
         fetch-depth: 0
 
@@ -34,7 +34,7 @@ jobs:
 
     steps:
     - name: Checkout repository
-      uses: actions/checkout@v4
+      uses: actions/checkout@v5
 
     - name: Set up Go
       uses: actions/setup-go@v5
--- a/helm/cluster/tests/go.mod
+++ b/helm/cluster/tests/go.mod
@@ -2,7 +2,7 @@ module github.com/giantswarm/cluster/helm/cluster/tests
 
 go 1.23.0
 
-toolchain go1.24.6
+toolchain go1.25.0
 
 require (
 	github.com/onsi/ginkgo/v2 v2.23.4


Gacko (changes requested):
Please make it "enabled" and true by default - thank you!

anvddriesch:
> What about the reference to the KubeadmConfig in the MachinePool resource?

In case of eks we set a custom bootstrap template using `providerIntegration.resourcesApi.machinePool.bootstrap.templateName` instead.

> Please make it "enabled" and true by default - thank you!

Sure, if you prefer that. :+1:

AverageMarcus:
> In case of eks we set a custom bootstrap template using providerIntegration.resourcesApi.machinePool.bootstrap.templateName instead.

Should we add a guard of some sort in case this is accidentally set in the future? E.g. within the MachinePool block add a check for that `enabled: true` and if not it errors? Hopefully it'll never be needed though.

Gacko:
> Sure, if you prefer that. 👍

Well, I thought it's a common best practice. 😄

anvddriesch:
> Should we add a guard of some sort in case this is accidentally set in the future? E.g. within the MachinePool block add a check for that enabled: true and if not it errors? Hopefully it'll never be needed though.

We could also separate the ref field and only render it if the kubeadmconfig is enabled?

anvddriesch:
> I think that'll break the resource. Fairly sure it's a required field.
I don't think the reference itself is required since some providers are not using it (like eks for example) but I guess `bootstrap` can not be empty.

CHANGELOG.md
```diff
@@ -7,6 +7,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Added
+
+- Add `disable` flag to `providerIntegration.workers.kubeadmConfig` to disable the rendering of the `KubeadmConfig` resource.
```
Gacko, line 12:
```suggestion
- Add `enabled` flag to `providerIntegration.workers.kubeadmConfig` to disable the rendering of the `KubeadmConfig` resource.
```

helm/cluster/README.md
```diff
@@ -949,6 +949,7 @@ Provider-specific properties that can be set by cluster-$provider chart in order
 | `providerIntegration.workers.defaultNodePools.PATTERN.replicas` | **Replicas** - The number of node pool nodes.|**Type:** `integer`<br/>**Key pattern:**<br/>`PATTERN`=`^[a-z0-9][-a-z0-9]{3,18}[a-z0-9]$`<br/>|
 | `providerIntegration.workers.defaultNodePools.PATTERN.type` | **Node pool type** - Controller that will manage the node pool. Only used to enable karpenter node pools.|**Type:** `string`<br/>**Key pattern:**<br/>`PATTERN`=`^[a-z0-9][-a-z0-9]{3,18}[a-z0-9]$`<br/>**Default:** `"machinepool"`|
 | `providerIntegration.workers.kubeadmConfig` | **Kubeadm config** - Configuration of workers nodes.|**Type:** `object`<br/>|
+| `providerIntegration.workers.kubeadmConfig.enable` | **Enable kubeadm config** - Flag that indicates if the kubeadm config should be rendered.|**Type:** `boolean`<br/>**Default:** `true`|
```
Gacko, line 952:
```suggestion
| `providerIntegration.workers.kubeadmConfig.enabled` | **Enable kubeadm config** - Flag that indicates if the kubeadm config should be rendered.|**Type:** `boolean`<br/>**Default:** `true`|
```

helm/cluster/templates/clusterapi/workers/kubeadmconfig.yaml
```diff
@@ -12,6 +12,7 @@ node rollouts and thus possibly interruped workloads for customers.
 */}}
 {{- if $.Values.providerIntegration.resourcesApi.machinePoolResourcesEnabled }}
 {{- if eq $.Values.providerIntegration.resourcesApi.nodePoolKind "MachinePool" }}
+{{- if $.Values.providerIntegration.workers.kubeadmConfig.enable }}
```
Gacko, line 15:
```suggestion
{{- if $.Values.providerIntegration.workers.kubeadmConfig.enabled }}
```

helm/cluster/values.schema.json
```diff
@@ -3118,6 +3118,12 @@
                             ],
                             "additionalProperties": false,
                             "properties": {
+                                "enable": {
```
Gacko, line 3121:
```suggestion
                                "enabled": {
```

helm/cluster/values.yaml
```diff
@@ -237,6 +237,7 @@ providerIntegration:
   useReleases: false
   workers:
     kubeadmConfig:
+      enable: true
```
Gacko, line 240:
```suggestion
      enabled: true
```

Gacko (changes requested):

PR #613, opened 2025-09-03
Chart: Obsolete default config with mirroring for Docker Hub registry.

piontec:
For a long time, docker hub shouldn't be used at all to pull our images: they introduced harsh rate limiting and high prices to increase them, so we switched to `gsoci.azurecr.io` entirely. If any image from dockerhub is needed, it should be first copied to our registry using `retagger` and then pulled from `gsoci.azurecr.io`. As such, we should remove any mirroring for docker hub. Also, the whole `giantswarm.azurecr.io` registry is considered legacy and will be deleted in the future.

Someone more knowledgeable about cluster deployments: please review this, taking into scope the impact on containerd's configuration and a risk that some cluster images might still come from docker hub.

Related - [issue tracking last images not pulled from gsoci](https://github.com/giantswarm/giantswarm/issues/34097)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,10 +2,14 @@
 
 All notable changes to this project will be documented in this file.
 
-The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
-and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
+The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/), and this project adheres to
+[Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
 ## [Unreleased]
+
+### Removed
+
+- Chart: obsolete default config with mirroring for docker hub registry
 
 ## [2.6.2] - 2025-09-02
 
@@ -23,7 +27,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Added
 
-- Add `enabled` flag to `providerIntegration.workers.kubeadmConfig` to disable the rendering of the `KubeadmConfig` resource.
+- Add `enabled` flag to `providerIntegration.workers.kubeadmConfig` to disable the rendering of the
+  `KubeadmConfig` resource.
 - containerd: Add flag to enable SELinux.
 
 ## [2.5.0] - 2025-07-31
@@ -37,7 +42,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Changed
 
-- Introduce `type` property in the node pools to be consumed by provider charts when they want to use `karpenter`.
+- Introduce `type` property in the node pools to be consumed by provider charts when they want to use
+  `karpenter`.
 
 ## [2.3.0] - 2025-05-15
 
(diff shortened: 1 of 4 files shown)


Gacko:
Do I understand correctly, that with this config, an image referenced as `docker.io` would in fact being pulled from the mirror URL, while without this config we just use `docker.io` and therefore could get rate-limited? So worst case: We get rate-limited, but image pulling still works. Right?

Wouldn't it therefore make sense to have `gsoci.azurecr.io` as an endpoint for `docker.io` instead? I know this is not a nice solution, but I'm afraid of the impact of simply removing it while we are obviously still pulling images via `docker.io`.

Gacko (changes requested):
Do I understand correctly, that with this config, an image referenced as `docker.io` would in fact being pulled from the mirror URL, while without this config we just use `docker.io` and therefore could get rate-limited? So worst case: We get rate-limited, but image pulling still works. Right?

Wouldn't it therefore make sense to have `gsoci.azurecr.io` as an endpoint for `docker.io` instead? I know this is not a nice solution, but I'm afraid of the impact of simply removing it while we are obviously still pulling images via `docker.io`.

piontec:
> Do I understand correctly, that with this config, an image referenced as `docker.io` would in fact being pulled from the mirror URL, while without this config we just use `docker.io` and therefore could get rate-limited? So worst case: We get rate-limited, but image pulling still works. Right?

We don't and shouldn't pull anything from docker.io. We paid them for the non-rate-limited account and they played as badly and hence the whole migration to azurecr. We're not getting back to docker, don't use it. Now, this config has a more side effect: any time anything (think about customer's workloads as well) wants to go to dockerhub, we rewrite it to go to `giantswarm.azurecr.io` - TBH, I'm surprised it works this well, `containerd` must be handling failover really good.

> Wouldn't it therefore make sense to have `gsoci.azurecr.io` as an endpoint for `docker.io` instead? I know this is not a nice solution, but I'm afraid of the impact of simply removing it while we are obviously still pulling images via `docker.io`.

No, for the reasons above: we shouldn't use docker hub at all and we really don't want to override every customers' usage of docker hub to our gsoci.

TL;DR: don't use docker hub directly, please - that way we don't suffer because of their request limits and we let customers to provide and use (potentially) their own imagePullSecrets without doing hidden magic of redirectnig it elsewhere.

Gacko:
I think there is a misunderstanding.

Currently and without your change `docker.io/some/image:v1.0.0` would be pulled from `registry-1.docker.io/some/image:v1.0.0` and, if this fails or it does not exist there, `giantswarm.azurecr.io/some/image:v1.0.0`.

With your change `docker.io/some/image:v1.0.0` would be pulled directly from `docker.io`. So your change does not change where an image is being pulled from because this is being defined in the actual apps, not in containerd.

I totally get your point. We should migrate our apps to use `gsoci.azurecr.io` instead of `docker.io`. We can do this, sure. But it will take time and in the meantime we could already provide a solution that automatically does that under the hood by using something like the following:

```
                                        "docker.io": [
                                            {
                                                "endpoint": "gsoci.azurecr.io"
                                            },
                                            {
                                                "endpoint": "registry-1.docker.io"
                                            },
                                        ],
```

Also I wonder why we are defining the following:

```
                                        "gsoci.azurecr.io": [
                                            {
                                                "endpoint": "gsoci.azurecr.io"
                                            }
                                        ]
```

Isn't this redundant?

PR #625, opened 2025-09-11
Use helmrelease for the karpenter-bundle

fiunchinho:
### What does this PR do?

Towards https://github.com/giantswarm/cluster-aws/pull/1422

It takes around 15m to get workers relying on the app platform to deploy karpenter. Deploying karpenter using a `HelmRelease`, it takes 5m. This has been manually tested.


njuettner (approved):

fiunchinho:
I thought I needed it in both schemas. I agree with what you said. If it's not required, then I'll remove it and release a new patch.

Gacko:
Theoretically you do not need it in both schemas - or at least we should not have it in `cluster` schema. But I haven't checked if we allow additional properties here (I am on my phone right now). So yeah, if possible it would be nice to only have it in the consuming chart and maybe, if required, allow additional properties.

fiunchinho:
I created this one https://github.com/giantswarm/cluster/pull/633

PR #626, opened 2025-09-11
feat(cluster/teleport): teleport-auto-update service

ssyno:
### What does this PR do?

Towards https://github.com/giantswarm/giantswarm/issues/33205

### How does it look like?


stone-z (commented):
This is already reasonable at face value but here some initial thoughts.

In addition:
- I'd suggest including a `defaultVersion` key which can be pinned in a release. That way, when a new cluster starts, unless `manualVersionOverride` is set, it will get `defaultVersion` and then be upgraded to the new target version after the first update run. But the default version allows us to more directly version Teleport with releases by disabling auto-update if we want.
- I couldn't find any public signatures for the linux binaries, but if they exist it would be good to include in the verification.
- What do you think about [this](https://goteleport.com/docs/upgrading/agent-managed-updates/)?

AndiDog (commented):
If the automated update goes wrong, we lose access. Did we consider other options?

PR #629, opened 2025-09-16
Remove capi-node-labeler app

fiunchinho:
### What does this PR do?

Towards https://github.com/giantswarm/giantswarm/issues/33938

From now on, the worker nodes won't have the `node-role.kubernetes.io/worker` or `node.kubernetes.io/worker` labels.

We've checked all WCs and found no customers relying on these worker labels, so we've decided to remove them. Our own apps using them have been changed already to use alternative methods when needed.

We are planning on releasing this as a new major release of the cluster chart and the provider charts, and including those in v33.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,7 +7,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
-### Changed
+### Removed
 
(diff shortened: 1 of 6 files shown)


njuettner (approved):

fiunchinho:
> LGTM! I assume you also tested this in `cluster-aws`.

I manually created a cluster. But e2e are also currently running in here https://tekton.ci.giantswarm.io/#/namespaces/tekton-pipelines/pipelineruns/pr-cluster-aws-1432-cluster-test-suitest7cj4?pipelineTask=create-github-check&view=logs

PR #636, opened 2025-09-25
Chart: Reduce Cilium reconciliation interval to 1 minute.

Gacko:
Currently the Cilium HelmRelease CR is only being reconciled every 5 minutes (hard-coded default value in the Helm Releases templating).

Since this CR is being created at the same time as the Cluster and Infrastructure CRs, it might happen Cilium gets first installed after 5 minutes, which would be ok, but often it happens to take 10 minutes or even 15 minutes. In general a reconciliation (and with that retry) interval of 5 minutes seems quite high for such an important core component. Keep in mind: Nodes are not getting ready without the CNI, so you cannot install anything else anyway as long you are not ignoring these taints.

Therefore I'd ask to reduce it to 1 minute with this change.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Changed
+
+- Chart: Reduce Cilium reconciliation interval to 1 minute.
+
 ## [4.0.1] - 2025-09-24
 
 ### Fixed
--- a/helm/cluster/files/helmreleases/cilium.yaml
+++ b/helm/cluster/files/helmreleases/cilium.yaml
@@ -4,6 +4,7 @@ namespace: kube-system
 # used by renovate
 # repo: giantswarm/cilium-app
 version: 1.3.1
+interval: 1m
 timeout: 1h
 remediationRetries: 0
 defaultValues:


Gacko:
Actually I do not know a lot about the consequences of this change. I only often see E2E clusters taking long to come up. So for real and long running clusters this most likely does not change a lot, but also I do not know if it is a good idea to add a possibility to override it here and then only override it for E2E clusters, because that then makes our tests less meaningful.

So overall I would be fine with both -m erging or not merging this - and am open for any kind of concerns or suggestions.

Gacko:
> since there are other components that will also reconcile at 5mins

There are also other components reconciling at 1 minute, like the Vertical Pod Autoscaler CRDs, which are a dependency for many other components. So overall we already go in the direction of having dependencies reconcile earlier. Sadly there is no way to differentiate between first installation and reconciliation in Helm Releases.

Also these other components reconciling at 5 minutes might depend on Cilium and therefore get delayed by more than 5 minutes in the end.

Gacko:
From [here](https://github.com/giantswarm/cluster/pull/636#issuecomment-3333306209):

> but also I do not know if it is a good idea to add a possibility to override it here and then only override it for E2E clusters in the cluster standup/teardown component, because that then makes our tests less meaningful.

😅

mcharriere:
I think we should ask @giantswarm/team-honeybadger , as they own the component that could be impacted by this change. 1m is the minimum value.

Also, I can't find metrics that support your assumptions. 
```
gotk_resource_info{app="kube-state-metrics", name=~"t-.*-cilium",ready="False"}
```
Running this query in grizzly I don't see many taking longer than 8mins

mcharriere:
my concern is mainly about is that this change conditions the resource usage of the helm-controller and is too buried in the cluster chat that makes it hard to change in the future.
I can imagine that in MCs with a high number of WCs, the helm-controller will struggle to keep with all the helmreleases

Gacko:
It seems I am not the only one considering that: https://github.com/giantswarm/cluster/tree/cilium-install-quicker

@AndiDog, in case you wanna join the discussion or get early feedback on what you are about to do.

Gacko:
Hm, I do not think it's worth the effort hacking something into `cluster-standup-teardown` just for that. I would like to keep stuff close to production, especially if the root cause here is something different that would also improve production if fixed properly.

PR #657, opened 2025-10-07
Add shared helper functions with ConfigMap for Cluster Apps lookup

njuettner:
Towards: https://github.com/giantswarm/giantswarm/issues/34176

Adds three new shared helper functions to the `cluster` chart that can be called from provider charts (cluster-aws, cluster-vsphere, etc.):

- `cluster.connectivity.baseDomain` - Returns base domain
- `cluster.managementCluster` - Returns management cluster name
- `cluster.components.containerd.containerRegistries` - Returns container registry configuration
- `cluster.internal.get-config-map-values` - Internal helper that caches ConfigMap lookup

## Why

This eliminates code duplication across provider charts and provides a centralized way to configure cluster values via the `cluster-app-config` ConfigMap in the `default` namespace.

## How It Works

1. Look up value from `cluster-app-config` ConfigMap in `default` namespace
2. If not found, fallback to Helm values (backward compatibility)
3. ConfigMap takes precedence when both exist

**ConfigMap Format:**
```yaml
apiVersion: v1
kind: ConfigMap
metadata:
  name: cluster-app-config
  namespace: default
data:
  values.yaml: |
    baseDomain: capi.pek.aws.k8s.adidas.com.cn
    managementCluster: galaxy
    containerRegistries:
      gsoci.azurecr.io:
        - endpoint: giantswarm-registry.cn-shanghai.cr.aliyuncs.com
        - endpoint: gsoci.azurecr.io
```

Can be called in provider charts like `
{{- $baseDomain := include "cluster.connectivity.baseDomain" $ -}}`

Tested it manually in `golem` by applying a configmap similar to https://github.com/giantswarm/giantswarm-management-clusters/pull/1426.

```
apiVersion: v1
data:
  values.yaml: |
    baseDomain: nick.evilcorp.com
    managementCluster: fillermc
kind: ConfigMap
metadata:
  name: cluster-app-config
  namespace: default
```

Modified certain places in https://github.com/giantswarm/cluster-aws/pull/1480 where we use `baseDomain` in `cluster-aws`

Overwriting `baseDomain` in `IRSA Claim` as an example:

```
spec:
  domain: nick.nick.evilcorp.com
  name: nick
  providerConfigRef: nick
  region: eu-west-2
  resourceRef:
    apiVersion: crossplane.giantswarm.io/v1
    kind: IRSA
    name: nick-gngt8
```

This will also allow us to drop the webhook injection once we have transitioned.

### DIFF
--- a/helm/cluster/files/etc/containerd/config.toml
+++ b/helm/cluster/files/etc/containerd/config.toml
@@ -65,4 +65,3 @@ enable_selinux = {{ include "cluster.containerd.selinux.enabled" $ }}
   [plugins."io.containerd.grpc.v1.cri".containerd.runtimes.nvidia.options]
     BinaryName = "/usr/bin/nvidia-container-runtime"
     SystemdCgroup = true
-


helm/cluster/files/etc/containerd/config.toml
```diff
@@ -64,3 +65,4 @@ enable_selinux = {{ include "cluster.containerd.selinux.enabled" $ }}
   [plugins."io.containerd.grpc.v1.cri".containerd.runtimes.nvidia.options]
     BinaryName = "/usr/bin/nvidia-container-runtime"
     SystemdCgroup = true
+
```
Gacko, line 68:
```suggestion
```

PR #780, opened 2026-02-11
feat(chart): Migrate Apps to HelmReleases.

fiunchinho:
Towards https://github.com/giantswarm/roadmap/issues/4215

This pull request adds a `pre-update` Helm hook `Job` to migrate `App` CRs to `HelmReleases` CRs. The changes on this PR need to be released in the same GiantSwarm workload cluster Release than these other PRs (likely v35):

- https://github.com/giantswarm/aws-nth-bundle/pull/53
- https://github.com/giantswarm/observability-bundle/pull/368
- https://github.com/giantswarm/security-bundle/pull/654

The `pre-update` job could be removed on the next major GiantSwarm workload cluster Release (likely v36), once all clusters have gone through a release containing the migration hook.

This PR also adds a `pre-delete` Helm hook `Job` to clean up `HelmReleases` on cluster deletion. This is basically restoring what we deleted here https://github.com/giantswarm/cluster/pull/711
This is now required again, because from time to time, I've seen that `flux` does not have enough time to clean up all the `HelmReleases` that we have for the default apps before the cluster control plane gets deleted, leaving leftovers.

### Why do we need a migration hook?

We are migrating the default applications to be managed by **Flux HelmRelease** CRs instead of **Apps** CRs. The challenge is that both systems ultimately manage Helm releases (stored as secrets in the target cluster). If not handled carefully, both systems could fight over the same Helm release, causing failures, or the actual running applications could be uninstalled during the transition.

The migration hook runs as a Helm `pre-upgrade` hook — it executes **before** the new chart version (which contains HelmRelease CRs instead of App CRs) is applied. Its job is to cleanly remove the old App CRs and Chart CRs without disturbing the actual running applications, so that Flux can adopt the existing Helm releases seamlessly.

### Two types of default applications

- **Non-bundle apps** (e.g., `cilium`, `coredns`, `cert-manager`): One `App` CR in the Management Cluster (MC) → one `Chart` CR in the Workload Cluster (WC) → one Helm chart installed in the WC.
- **Bundle apps** (e.g., `security-bundle`, `observability-bundle`, `aws-nth-bundle`): One `App` CR in the MC → one `Chart` CR in the MC → when installed, creates multiple sub-`App` CRs → each sub-App creates a `Chart` CR in the WC → each installed in the WC.

This distinction is critical because the two types require different cleanup strategies. The Job uses label selectors to differentiate between bundle and non-bundle apps, but also between default and non default apps (apps installed  by the customer)

```
giantswarm.io/cluster=$CLUSTER   → Belongs to this specific cluster

release.giantswarm.io/version    → Present on all default apps (set by the cluster chart)

app-operator.giantswarm.io/version!=0.0.0   → Non-bundle apps (version 0.0.0 = bundle)
```

## Migration Phases

The migration hook operates in three phases, executed sequentially. Each phase has a specific purpose and handles a different category of resources.

### Phase 0: Remove Dependency Annotations

```
kubectl annotate apps -n "$NAMESPACE" \
  -l "giantswarm.io/cluster=$CLUSTER,release.giantswarm.io/version" \
  "app-operator.giantswarm.io/depends-on-"
```

**What it does:** Removes the `app-operator.giantswarm.io/depends-on` annotation from all default App CRs.

**Why it's needed:** Some App CRs declare dependencies on other App CRs (e.g., `kyverno-policies` depends on `kyverno-crds`). When `app-operator` sees a dependency being paused or deleted, it pauses the dependent's **Chart CR** in the MC. If we don't remove these annotations first, pausing App CRs in Phase 1 would cascade into pausing Chart CRs that we don't want to pause yet — creating unpredictable ordering issues.

By stripping the annotations upfront, each App CR becomes independent and can be handled in isolation.

### Phase 1: Non-bundle Apps + WC Chart CRs

This phase handles all non-bundle default applications. The goal is to remove both the App CRs (in the MC) and the Chart CRs (in the WC) without triggering `helm uninstall` on the actual applications.

#### Phase 1a: Pause non-bundle App CRs

```
kubectl annotate apps -n "$NAMESPACE" \
  -l "$NON_BUNDLE_SELECTOR" \
  "operatorkit.giantswarm.io/paused=true" --overwrite
```

**What it does:** Adds the `operatorkit.giantswarm.io/paused=true` annotation to all non-bundle default App CRs.

**Why it's needed:** When `app-operator` is paused on an App CR, it stops reconciling it. This means it won't re-create the Chart CR in the WC if we delete it, and it won't interfere with our cleanup operations.

#### Phase 1b: Remove finalizers from non-bundle App CRs

```
for APP in $NON_BUNDLE_APPS; do
  kubectl patch apps "$APP" -n "$NAMESPACE" --type=json \
    -p='[{"op": "remove", "path": "/metadata/finalizers"}]' &
done
```

**What it does:** Strips `metadata.finalizers` from each non-bundle App CR using parallel batches of 10.

**Why it's needed:** `app-operator` adds a finalizer to App CRs. When an App CR with a finalizer is deleted, Kubernetes doesn't actually remove it — it marks it for deletion and waits for `app-operator` to process its cleanup logic (which includes running `helm uninstall` via the Chart CR). By removing the finalizer first, we ensure deletion is instant and `app-operator` never runs its cleanup.

**Why parallel batches:** With ~20 non-bundle apps, sequential patching would be slow. Parallel batches of 10 complete in seconds.

#### Phase 1c: Delete non-bundle App CRs

```
kubectl delete apps -n "$NAMESPACE" \
  -l "$NON_BUNDLE_SELECTOR" \
  --ignore-not-found=true --wait=false
```

**What it does:** Deletes all non-bundle default App CRs from the MC.

**Why `--wait=false`:** Since we already removed finalizers, deletion is instant. No need to wait.

**What happens after:** The App CRs are gone. `app-operator` is no longer watching them (they were paused anyway). The Chart CRs in the WC still exist, but no one will re-create them if we delete them — because the parent App CRs are gone.

#### Phase 1d: Clean up WC Chart CRs

This sub-phase connects to the WC using the cluster's kubeconfig and handles the Chart CRs there.

**Why we pause before removing:** `chart-operator` runs in the WC and watches Chart CRs. If a Chart CR is deleted without pausing first, `chart-operator` sees the deletion event and runs `helm uninstall` — which would remove the actual application. By pausing first, `chart-operator` ignores the Chart CR entirely.

**Why we remove finalizers:** Same reasoning as App CRs — `chart-operator` adds finalizers to Chart CRs. Without removing them, Kubernetes would wait for `chart-operator` to process the deletion (which means running `helm uninstall`).

**End result of Phase 1:** All non-bundle App CRs are gone from the MC. All non-bundle Chart CRs are gone from the WC. But the actual Helm releases (secrets like `sh.helm.release.v1.cilium.v1`) remain untouched in the WC. When Flux later creates HelmRelease CRs for these apps, it will discover the existing Helm releases and adopt them.

### Phase 2: Bundle App CRs via Operator Cascade

Bundles require a fundamentally different approach. A bundle's Chart CR lives in the `giantswarm` namespace in the **MC** (not the WC). When `chart-operator` installs the bundle chart, it creates sub-App CRs in the MC, which in turn create Chart CRs in the WC, which install the actual applications. The sub-App Chart CRs in the WC were already handled in Phase 1d. But the bundle's own Chart CR in the MC and its sub-App CRs need to be cleaned up too.

We **cannot** use the same pause-and-strip approach for bundles because:
1. The bundle's Chart CR is in the MC, and our RBAC doesn't allow patching Chart CRs in the MC (they live in the `giantswarm` namespace).
2. Even if we could, the bundle's sub-App CRs need to be cleaned up properly, and that cleanup logic lives in `app-operator`'s finalizer handling.

Instead, we **let the operators do their job**: we delete the bundle App CR and let `app-operator` cascade the deletion naturally.

#### Phase 2a: Delete bundle App CRs (keeping finalizers)

```
for APP in $REMAINING_APPS; do
  is_operator "$APP" && continue
  kubectl delete apps "$APP" -n "$NAMESPACE" \
    --ignore-not-found=true --wait=false
done
```

**What it does:** Deletes the remaining default App CRs (bundles) **without** stripping finalizers first.

**Why keep the finalizers:** We *want* `app-operator` to process its cleanup logic here. The bundle's finalizer handler in `app-operator` will cascade-delete the MC Chart CR, which triggers `chart-operator` to run `helm uninstall` on the bundle chart. This `helm uninstall` removes the sub-App CRs from the MC (the bundle chart's resources). The sub-App CRs' own Chart CRs in the WC were already neutralized in Phase 1d (paused, finalizers stripped, deleted), so no `helm uninstall` runs against the actual WC applications.

**What `is_operator` skips:** The `app-operator` and `chart-operator` App CRs are explicitly excluded — these operators must remain alive to process the cascade.

#### Phase 2b: Wait for cascade completion

```
while [ $ELAPSED -lt $TIMEOUT ]; do
  ALL_GONE=true
  for APP in $BUNDLE_APPS; do
    if kubectl get apps "$APP" -n "$NAMESPACE" >/dev/null 2>&1; then
      ALL_GONE=false
      break
    fi
  done
  if [ "$ALL_GONE" = "true" ]; then break; fi
  sleep 5
done
```

**What it does:** Polls every 5 seconds (up to 5 minutes) until all bundle App CRs have been fully deleted by the operator cascade.

**Why this wait is critical:** If the hook exits while the cascade is still running, Helm will apply the new chart version — which creates the new HelmRelease CRs. But the old cascade is still deleting things. The old `chart-operator` running `helm uninstall` on the bundle chart will delete the sub-App CRs, but Flux may have already started creating the **new** bundle sub-HRs. The old cascade could accidentally delete the newly-created HelmReleases, causing permanently missing applications.

By waiting for the cascade to finish, we ensure a clean handoff: all old App CRs are gone, then Helm applies the new chart with HelmRelease CRs, and Flux takes over.

## After the Hook: Flux Takes Over

Once the hook completes:

1. **Helm applies the new chart version** — this creates HelmRelease CRs instead of App CRs for all default applications.
2. **Flux discovers the HelmRelease CRs** and checks the target cluster for existing Helm releases.
3. **Flux adopts the existing Helm releases** — since the Helm release secrets are still present in the WC, Flux recognizes them and takes ownership. No reinstall needed.
4. **Subsequent reconciliations** are handled entirely by Flux, with no involvement from `app-operator` or `chart-operator`.


helm/cluster/templates/apps/helmreleases.yaml
```diff
@@ -159,11 +154,17 @@ spec:
   upgrade:
     disableWait: {{ $app.disableWait | default false }}
     remediation:
+      remediateLastFailure: true
+      strategy: uninstall
```
AndiDog: Sounds a little dangerous compared to the default strategy (`rollback`). Why choose this one?

helm/cluster/templates/apps/app-to-hr-migration-hook/job.yaml
```diff
@@ -0,0 +1,227 @@
...
+          # ============================================================
+          # Phase 0: Remove dependency annotations
+          # ============================================================
+          # Remove depends-on annotations from default App CRs.
+          # This prevents app-operator from pausing MC Chart CRs when
+          # we later pause/delete the dependency App CRs in Phase 1.
+          echo "Phase 0: Removing depends-on annotations from default App CRs"
+          kubectl annotate apps.application.giantswarm.io -n "$NAMESPACE" \
+            -l "$CLUSTER_SELECTOR" \
+            "app-operator.giantswarm.io/depends-on-" 2>/dev/null || true
```
AndiDog: [repeated]

Better don't swallow error messages. Since removal of a non-existent counts as success, we should anyway only see hard errors here, I think.

Also, is `|| true` reasonable? For example, if all commands in the script fail, but failures are ignored, the hook will still succeed which is probably not what we want?!

helm/cluster/templates/apps/app-to-hr-migration-hook/job.yaml
```diff
@@ -0,0 +1,227 @@
...
+              done
+              if [ "$ALL_GONE" = "true" ]; then
+                echo "All bundle App CRs fully deleted after ${ELAPSED}s"
+                break
+              fi
+              sleep 5
+              ELAPSED=$((ELAPSED + 5))
+            done
+            if [ "$ALL_GONE" != "true" ]; then
+              echo "WARNING: Timed out waiting for bundle App CRs to be deleted after ${TIMEOUT}s"
```
AndiDog: Shouldn't this be a failure so we notice the error at all by having the hook retry?

helm/cluster/templates/apps/helmreleases.yaml
fiunchinho: While running the migration, sometimes HelmReleases fail to install due to tiemouts, throttling, etc. And if the installation fails, the `helm-controller` will block the HelmRelease because there is no previous release to rollback to, requiring manual intervention to unblock and retry the installation. To avoid manual intervention, I tried the `uninstall` strategy. This is what I wrote in my notes

> When Flux adopts an existing Helm release during migration, there's no prior Flux-managed revision to roll back to. If the first upgrade after adoption fails (e.g., context-deadline-exceeded), Flux tries to rollback but hits `MissingRollbackTarget` and stalls permanently. The HR stays in `UpgradeFailed` forever, requiring manual intervention.

I can't remember more specifics because it's been a while since I worked on this. I'll try to find some time to re-test the whole thing again with a less destructive remediation strategy, but I guess I'll hit the same issues.

helm/cluster/templates/apps/app-to-hr-migration-hook/job.yaml
fiunchinho: yeah, good point. I'll make the migration fail whenever there is an error.

helm/cluster/templates/apps/helmreleases.yaml
fiunchinho: After your comment, I started checking if applications in the WC were being uninstalled as part of the process, and apparently they did, even when not using `strategy: uninstall` in the HelmReleases. I obsessed over this the whole last week, trying different approaches to avoid uninstalling applications in the WC, and when I was about to give up, I manage to get it working. So the hook is a bit different now. Also it's not using `remediation.strategy: uninstall` anymore. The problem regarding `remediation.strategy: uninstall` is that _sometimes_ there would be an error while flux is adopting the applications (i.e. transient timeout), and if this happens, there is no previous helm release to roll back to, so the application will get stuck in an error state forever until manual intervention. After many tests, my opinion is that it does not happen frequently enough to make it worth the risk of adding `remediation.strategy: uninstall` to all HelmReleases. So I got rid of `remediation.strategy: uninstall`, and we need to live with the fact that _sometimes_ we may need to `suspend / unsuspend` a HelmRelease manually during the migration.

helm/cluster/templates/apps/app-to-hr-migration-hook/job.yaml
```diff
@@ -76,152 +94,477 @@ spec:
...
+                cat "$errfile" >&2
+                sleep $((i * 2))
+                i=$((i + 1))
+                continue
+              fi
+              cat "$errfile" >&2
+              rm -f "$errfile"
+              return $rc
+            done
+            echo "  ERROR: kubectl $1 exhausted $max retries" >&2
```
AndiDog:
```suggestion
            echo "  ERROR: kubectl $1 exhausted $max attempts" >&2
```

5 attempts = 5 tries = 4 **re**tries [I'm being pedantic]

helm/cluster/templates/apps/app-to-hr-migration-hook/job.yaml
```diff
@@ -76,152 +94,477 @@ spec:
...
+            kubectl_retry delete apps.application.giantswarm.io -n "$NAMESPACE" \
+              -l "$SEL" \
+              --ignore-not-found=true --wait=false
+          }
+
+          # Delete WC Chart CRs by selector.
+          delete_charts_in_wc() {
+            DESC="$1"; SEL="$2"
+            echo "  Deleting $DESC"
+            kubectl_retry $WC delete chart -n "$CHART_NAMESPACE" \
```
AndiDog: Just to be sure, we could refer to the full CRD:

`s/(annotate|get|delete|patch) chart /\1 chart.application.giantswarm.io /`

helm/cluster/templates/apps/app-to-hr-migration-hook/job.yaml
```diff
@@ -1,28 +1,41 @@
...
+                            (rendered by the bundle helm chart) and
+                            do NOT have release.giantswarm.io/version
 
-Phase 1 (Non-bundle App CRs + WC Chart CRs):
-  - Pause non-bundle App CRs (excludes operators via label selector)
-  - Remove finalizers + delete non-bundle App CRs
-  - Once App CRs are gone, clean up default app WC Chart CRs safely
-    (no risk of app-operator re-creating them)
+Phases:
+  0:  Strip depends-on annotations from default App CRs.
```
AndiDog: `0a` (upgrade timeout increase) is missing

helm/cluster/templates/apps/app-to-hr-migration-hook/job.yaml
```diff
@@ -76,152 +94,477 @@ spec:
...
-          COUNT=0
-          for CHART in $CHARTS; do
-            (kubectl $WC patch chart "$CHART" -n "$CHART_NAMESPACE" --type=json \
-              -p='[{"op": "remove", "path": "/metadata/finalizers"}]' 2>/dev/null || true) &
-            COUNT=$((COUNT + 1))
-            if [ $((COUNT % BATCH_SIZE)) -eq 0 ]; then wait; fi
+          # Confirm no chart-operator pods remain. Replicas=0 prevents the
+          # ReplicaSet from re-creating; force-delete made the existing pods
+          # disappear immediately.
+          for i in 1 2 3 4 5; do
```
AndiDog: [btw, minor]

```suggestion
          for i in $(seq 1 5); do
```

helm/cluster/templates/apps/app-to-hr-migration-hook/job.yaml
```diff
@@ -76,152 +94,477 @@ spec:
...
+            PODS=$(kubectl $WC get pods -n "$CHART_NAMESPACE" -l app=chart-operator --no-headers 2>/dev/null | wc -l | tr -d ' ')
+            if [ "$PODS" -eq 0 ]; then
+              echo "  chart-operator on WC is down"
+              break
+            fi
+            echo "  still seeing $PODS chart-operator pod(s) (attempt $i)"
+            sleep 2
           done
-          wait
```
AndiDog: This should check after all attempts if all chart-operator pods are ultimately gone

helm/cluster/templates/apps/app-to-hr-migration-hook/job.yaml
```diff
@@ -76,152 +94,477 @@ spec:
...
+          # helm uninstall on the bundle release on MC) to fire in Phase D.
+          # Phase D blocks until the cascade completes, so the bundle helm
+          # release storage on MC is fully gone before the hook exits and
+          # before helm proceeds to apply the cluster-aws upgrade diff. When
+          # Flux helm-controller eventually picks up the new bundle HR, it
+          # sees no existing release in storage and runs a fresh `helm install`
+          # at the new bundle version — no takeOwnership upgrade, no race
+          # with chart-operator-MC reverting it.
+          # The bundle uninstall manifest (OLD chart, e.g. 2.6.0) contains
+          # sub-App CRs in `org-jose` only — those were already removed in
```
AndiDog: This needs a little rewording – what do `2.6.0` and `org-jose` mean here? The example is actually below and repeated, so maybe they should be the same.

helm/cluster/templates/apps/app-to-hr-migration-hook/job.yaml
```diff
@@ -76,152 +94,477 @@ spec:
...
+            REMAINING_NAMES=$(kubectl get apps.application.giantswarm.io -n "$NAMESPACE" \
+              -l "$BUNDLE_APP_SELECTOR" -o jsonpath='{.items[*].metadata.name}' 2>/dev/null)
+            echo "  $REMAINING bundle App CRs remaining at ${ELAPSED}/${BUNDLE_DRAIN_TIMEOUT}s: $REMAINING_NAMES"
+            sleep 10
+            ELAPSED=$((ELAPSED + 10))
+          done
+          if [ "$REMAINING" != "0" ]; then
+            echo "  WARNING: bundle App CRs did not drain within ${BUNDLE_DRAIN_TIMEOUT}s"
+            kubectl get apps.application.giantswarm.io -n "$NAMESPACE" -l "$BUNDLE_APP_SELECTOR" \
+              -o custom-columns="NAME:.metadata.name,FINALIZERS:.metadata.finalizers,DELETIONTS:.metadata.deletionTimestamp" 2>/dev/null || true
```
AndiDog: Should this be a fatal error?

helm/cluster/templates/apps/app-to-hr-migration-hook/job.yaml
```diff
@@ -76,152 +94,477 @@ spec:
...
+          # Now poll for the bundle helm release SECRETS to be gone. This is
+          # the load-bearing check: helm release secrets stay in storage until
+          # chart-operator-MC's helm uninstall fully completes. Without this
+          # wait, the hook exits while secrets are still in storage; helm-
+          # controller's takeOwnership upgrade then claims the release; later
+          # chart-operator-MC's uninstall finishes and wipes the secrets,
+          # which deletes the manifest items (sub-app HRs) and triggers
+          # helm-controller to uninstall the WC sub-app releases.
+          echo "  Polling bundle helm release secrets to confirm uninstall completion..."
+          while [ $ELAPSED -lt $BUNDLE_DRAIN_TIMEOUT ]; do
```
AndiDog: Reusing the existing, probably non-zero value of `ELAPSED` is okay, I guess?

helm/cluster/templates/apps/app-to-hr-migration-hook/job.yaml
```diff
@@ -76,152 +94,477 @@ spec:
...
+            if [ "$ALL_GONE" = "true" ]; then
+              echo "  All bundle helm release secrets gone after ${ELAPSED}s — uninstall fully complete"
+              break
             fi
+            echo "  Bundle helm release secrets still present at ${ELAPSED}/${BUNDLE_DRAIN_TIMEOUT}s:$STILL_PRESENT"
+            sleep 10
+            ELAPSED=$((ELAPSED + 10))
+          done
+          if [ "$ALL_GONE" != "true" ]; then
+            echo "  WARNING: bundle helm release secrets still present after ${BUNDLE_DRAIN_TIMEOUT}s"
```
AndiDog: Also here: should this be a hard error since it would mean that Flux may try takeover rather than `helm install`?

helm/cluster/templates/apps/helmreleases.yaml
```diff
@@ -9,11 +9,11 @@
 {{- $app := tpl (.Files.Get $path) $ | fromYaml }}
 {{- $appName := $app.appName }}
 
-{{- /* For hr installations targeting the MC, we want to add the cluster name as a prefix to the appName 
+{{- /* For hr installations targeting the MC, we want to add the cluster name as a prefix to the appName
 to avoid naming conflicts in case of multiple clusters deployed in the same namespace.
 */}}
 {{- $releaseName := $appName }}
-{{- if $app.prependClusterName }}
+{{- if $app.inCluster }}
```
AndiDog, line 16: We don't read the `prependClusterName` property anymore, and this key is left behind in some files. I guess this now renders differently. Let's please re-check if this is desired and doesn't produce name clashes across WCs.

fiunchinho: The three properties that we used in some helm releases files:
- prependClusterName
- skipKubeConfig
- serviceAccountName

They all converge into `inCluster`. When installing a HR "in the MC", we want to use the `automation` SA, we do not want to render the `kubeConfig` section of the HR and we use the cluster name as prefix for the helm release.

fiunchinho:
@AndiDog — addressed the review comments in commits `f997d51` and `a33c9f8`. Quick summary:

**Cosmetic / wording**
- Line 134: `exhausted N retries` → `exhausted N attempts`.
- Header docstring: added the missing `0a:` (lift helm upgrade timeout to 15m).
- Line 326: `for i in $(seq 1 5); do`.
- Line 359 comment: rewritten to drop testing-specific `2.6.0` / `org-jose` references in favour of the production semantics.

**Reliability**
- All `kubectl ... chart` callsites now use `chart.application.giantswarm.io` fully-qualified.
- Inline comment around the bundle helm-secret poll explaining why we reuse `ELAPSED` (single shared timeout budget for the App CR drain + secret-poll halves of the cascade).

**Error handling**
- Phase 0a App CR timeout patch no longer `|| true` — fails the hook so chart-operator-MC retries.
- Phase 0b chart-operator pod scale-down: hard post-condition added (`FATAL exit 1` if any pod lingers after the retry loop).
- Phase D bundle App CR drain timeout: `WARNING` → `FATAL exit 1`.
- Phase D bundle helm release secret drain timeout: `WARNING` → `FATAL exit 1`.
- Redundant `|| true` on the `--ignore-not-found=true` pod delete dropped.

njuettner:
Minor cleanup follow-up: now that the template renders the `OCIRepository` unconditionally (the `{{- if $app.useOCISource }}` guard was dropped in `helmreleases.yaml`), `useOCISource: true` is dead config but is still present in three values files:

- `helm/cluster/files/helmreleases/cluster-probes.yaml`
- `helm/cluster/files/helmreleases/cluster-autoscaler-crossplane-resources.yaml`
- `helm/cluster/files/helmreleases/external-dns-crossplane-resources.yaml`

Harmless (the template no longer reads it), but worth removing for clarity.

fiunchinho:
> Minor cleanup follow-up: now that the template renders the `OCIRepository` unconditionally (the `{{- if $app.useOCISource }}` guard was dropped in `helmreleases.yaml`), `useOCISource: true` is dead config but is still present in three values files:
> 
> * `helm/cluster/files/helmreleases/cluster-probes.yaml`
> * `helm/cluster/files/helmreleases/cluster-autoscaler-crossplane-resources.yaml`
> * `helm/cluster/files/helmreleases/external-dns-crossplane-resources.yaml`
> 
> Harmless (the template no longer reads it), but worth removing for clarity.

Done! Good catch.

njuettner (approved):

Gacko (changes requested):
Please gimme a chance to review this until end of the week. Thank you!

helm/cluster/files/helmreleases/cert-manager.yaml
```diff
@@ -1,10 +1,9 @@
 appName: cert-manager
 chartName: cert-manager-app
 configKey: certManager
-clusterValues:
-  configMap: true
-  secret: true
-dependsOn: prometheus-operator-crd
+dependsOn:
+  - prometheus-operator-crd
+  - kube-prometheus-stack
```
Gacko: May I ask why you're using `kube-prometheus-stack` here? I imagine you added it because of the PodLogs CRs. If that’s the case, then `alloy-podlogs-crds` would be a better choice:

- https://github.com/giantswarm/observability-bundle/blob/main/helm/observability-bundle/values.yaml#L147-L155
- https://github.com/giantswarm/alloy-app/tree/main/helm

To be fair: It's a quite new chart in the Observability Bundle and `cert-manager` started to contain PodLogs CRs just 3 months ago (CAPI v34.1.0 Releases): https://github.com/giantswarm/releases/blob/master/capa/v34.1.0/release.yaml#L25-L29.

helm/cluster/files/helmreleases/chart-operator-extensions.yaml
```diff
@@ -1,10 +1,10 @@
 appName: chart-operator-extensions
 configKey: chartOperatorExtensions
-clusterValues:
-  configMap: false
-  secret: false
-dependsOn: prometheus-operator-crd
+dependsOn:
+- prometheus-operator-crd
+- kyverno-crds
```
Gacko: How comes this depends on Kyverno? I cannot see any Kyverno CRs in the chart: https://github.com/search?q=repo%3Agiantswarm%2Fchart-operator-extensions%20Kyverno&type=code.

helm/cluster/files/helmreleases/cluster-autoscaler.yaml
```diff
@@ -0,0 +1,19 @@
+appName: cluster-autoscaler
+chartName: cluster-autoscaler-app
+configKey: clusterAutoscaler
+inCluster: {{ eq $.Values.providerIntegration.provider "azure" }}
+dependsOn: {{ if not (eq $.Values.providerIntegration.provider "azure") }}["kyverno-crds"]{{ else }}[]{{ end }}
```
Gacko: Not sure if this works with the current code, but for readability reasons we could also use something like this here:


```suggestion
{{- if ne $.Values.providerIntegration.provider "azure" }}
dependsOn:
- kyverno-crds
{{- end }}
```

Also mind I replaced `not (eq ... )` with `ne`.

helm/cluster/files/helmreleases/external-dns.yaml
```diff
@@ -19,3 +16,12 @@ defaultValues:
...
     enabled: true
+  {{- if .Values.global.connectivity.proxy.enabled }}
+  env:
+    - name: NO_PROXY
+      value: {{ include "cluster.connectivity.proxy.noProxy" (dict "global" .Values.global "providerIntegration" .Values.providerIntegration) | quote }}
+    - name: HTTP_PROXY
+      value: {{ .Values.global.connectivity.proxy.httpProxy | quote }}
+    - name: HTTPS_PROXY
+      value: {{ .Values.global.connectivity.proxy.httpsProxy | quote }}
+  {{- end }}
```
Gacko, line 29: Just a nit, but since you are not indenting list items elsewhere:


```suggestion
  {{- if .Values.global.connectivity.proxy.enabled }}
  env:
  - name: NO_PROXY
    value: {{ include "cluster.connectivity.proxy.noProxy" (dict "global" .Values.global "providerIntegration" .Values.providerIntegration) | quote }}
  - name: HTTP_PROXY
    value: {{ .Values.global.connectivity.proxy.httpProxy | quote }}
  - name: HTTPS_PROXY
    value: {{ .Values.global.connectivity.proxy.httpsProxy | quote }}
  {{- end }}
```

helm/cluster/files/helmreleases/teleport-kube-agent.yaml
```diff
@@ -1,14 +1,19 @@
...
 # repo: giantswarm/teleport-kube-agent-app
 version: 0.11.1
+defaultValues:
+  {{- if .Values.global.connectivity.proxy.enabled }}
+  cluster:
+    proxy:
+      noProxy: {{ include "cluster.connectivity.proxy.noProxy" (dict "global" .Values.global "providerIntegration" .Values.providerIntegration) | quote }}
+      http: {{ .Values.global.connectivity.proxy.httpProxy | quote }}
+      https: {{ .Values.global.connectivity.proxy.httpsProxy | quote }}
+  {{- end }}
```
Gacko, line 15: These have changed in the latest version of Teleport Kube Agent, all the properties now live under a `teleport-kube-agent` top-level property since Shield started to make use of an upstream chart.

helm/cluster/templates/apps/app-to-hr-migration-hook/job.yaml
```diff
@@ -0,0 +1,728 @@
...
+  template:
+    metadata:
+      labels:
+        {{- include "cluster.labels.common.all" $ | nindent 8 }}
+    spec:
+      serviceAccountName: {{ include "cluster.resource.name" $ }}-migrate-apps
+      restartPolicy: Never
+      containers:
+      - name: kubectl
+        image: gsoci.azurecr.io/giantswarm/kubectl:{{ .Values.providerIntegration.kubernetesVersion }}
```
Gacko: This is v1.25.16, so heavily outdated. I'd rather rely on the Kubernetes version from the Release CR or update the default value to some more recent version.

helm/cluster/templates/apps/app-to-hr-migration-hook/job.yaml
```diff
@@ -0,0 +1,728 @@
...
+          requests:
+            cpu: 10m
+            memory: 128Mi
+          limits:
+            cpu: 200m
+            memory: 512Mi
+        command:
+        - /bin/sh
+        - -c
+        - |
```
Gacko:
```suggestion
        command:
        - /bin/sh
        args:
        - -c
        - |
```

helm/cluster/templates/apps/app-to-hr-migration-hook/job.yaml
```diff
@@ -0,0 +1,728 @@
...
+            cat "$errfile" >&2
+            rm -f "$errfile"
+            return 1
+          }
+
+          # Fetch WC kubeconfig
+          echo "Fetching WC kubeconfig for $CLUSTER_NAME"
+          kubectl_retry get secret -n "$NAMESPACE" "${CLUSTER_NAME}-kubeconfig" \
+            -o jsonpath='{.data.value}' | base64 -d > /tmp/wc-kubeconfig
+          WC="--kubeconfig /tmp/wc-kubeconfig"
```
Gacko: Wouldn't it make more sense to have this injected via a mount?

Gacko (changes requested):
Just a first round of review, not completely done, yet.

helm/cluster/files/helmreleases/cert-manager.yaml
fiunchinho: image


I created the `alloy-podlogs-crds` because of this task. It seems I forgotten to update the dependency here, good catch!

helm/cluster/files/helmreleases/chart-operator-extensions.yaml
fiunchinho: yeah, I don't think it's needed. I'll get rid of it.

helm/cluster/files/helmreleases/teleport-kube-agent.yaml
fiunchinho: Wait, I'm a bit confused. Isn't the `proxy` values the exception to this due to your own PR https://github.com/giantswarm/teleport-kube-agent-app/pull/158 included in 0.11.1? I think I may be miss understanding how it works. I understand the code on this PR is correct because proxy values do not live under a `teleport-kube-agent` top-level property.

Gacko: Ah, crap, my fault! I only reminded they moved things and there was some issue elsewhere, but the proxy we fixed this way. You're right, never mind my comment. Actually there've ben several issues with TKA lately, so one can get confused with it easily...

helm/cluster/templates/apps/app-to-hr-migration-hook/job.yaml
fiunchinho: bumped

helm/cluster/files/helmreleases/teleport-kube-agent.yaml
Gacko: One addition on that: We couldn't move the proxy variables to the `.teleport-kube-agent` key because they were provided via the cluster values and there we cannot respect each and every app. But with this PR here we could theoretically move the proxy variables to the `.teleport-kube-agent` key somewhen in the future, if that makes it easier for Shield.

Gacko:
One very important thing, before I forget about it again: We currently have two different definitions for Apps and HelmReleases in the `values.schema.json`:

- App: https://github.com/giantswarm/cluster/blob/migrate-apps-helmreleases/helm/cluster/values.schema.json#L14-L59
- HelmRelease: https://github.com/giantswarm/cluster/blob/migrate-apps-helmreleases/helm/cluster/values.schema.json#L354-L397

Currently your PR is not migrating these, especially because the HelmRelease definition does not support a `priority` property, yet, which is required for adapting the previous App CR behavior.

Second customers might have provided custom configuration for default apps in their clusters. Adapting the App Operator behavior for ConfigMap and Secret merging to HelmReleases might break their deployments, because currently their configuration is getting merged in the order they provide the configuration in the cluster values, but with this change the priority comes into play and might change the resulting values.

And last but not least: Have you checked if the currently used selectors for App & HelmRelease CRs might also affect customer-installed resources?

fiunchinho:
> Second customers might have provided custom configuration for default apps in their clusters. Adapting the App Operator behavior for ConfigMap and Secret merging to HelmReleases might break their deployments, because currently their configuration is getting merged in the order they provide the configuration in the cluster values, but with this change the priority comes into play and might change the resulting values.

Isn't this addressed by the PR just merged into this PR's branch? https://github.com/giantswarm/cluster/pull/909

Gacko:
I checked the diff of this PR and it only changes the default value of the Kubernetes version in the `values.schema.json`.

Gacko:
I rebased this PR on top of `main` and removed some duplicates / not customer-relevant details from the changelog. Also I aligned the way we truncate non-alphanumeric characters at the end of the chart-version template to how we (Tenet) implemented it elsewhere.

Here's the diff: https://github.com/giantswarm/cluster/compare/8f17739a07d8ea59e8524f11ce07818bd8254334..3906be6a649602b7ece5f315d6a628cdce02d075.

helm/cluster/templates/apps/helmreleases.yaml
```diff
@@ -159,11 +154,12 @@ spec:
   upgrade:
     disableWait: {{ $app.disableWait | default false }}
     remediation:
-      {{- if hasKey $app "remediationRetries" }}
-      retries: {{ $app.remediationRetries }}
-      {{- else }}
       retries: -1
-      {{- end }}
```
Gacko, line 166: Can you elaborate why you removed this? Cilium is making use of that, so maybe @cabbage can give us some insights if they also need it for upgrades.

helm/cluster/templates/apps/helmreleases.yaml
```diff
@@ -181,19 +177,48 @@ spec:
...
   {{- if $appHelmValues }}
-  values: {{- $appHelmValues | toYaml | nindent 4 }}
+  {{- $userValuesConfigMap = printf "%s-%s-user-values" (include "cluster.resource.name" $) $appName }}
   {{- end }}
-  {{- $extraConfigs := concat ($app.extraConfigs | default list) ($userConfig.extraConfigs | default list) }}
-  {{- if $extraConfigs }}
+  {{- $sortArgs := dict
+        "extraConfigs" $extraConfigs
+        "clusterValuesConfigMap" (ternary (printf "%s-cluster-values" (include "cluster.resource.name" $)) "" (not (empty ($app.clusterValues).configMap)))
+        "clusterValuesSecret" (ternary (printf "%s-cluster-values" (include "cluster.resource.name" $)) "" (not (empty ($app.clusterValues).secret)))
```
Gacko: Didn't you obsolete these from the HelmRelease definition files? And even if it was set, it would be a boolean, right? So shouldn't this be something like...

```
"clusterValuesConfigMap" (ternary (printf "%s-cluster-values" (include "cluster.resource.name" $)) "" ($app.clusterValues).configMap)
```

... instead?

Gacko: Overall it would be more readable to write this as:


```suggestion
        "clusterValuesConfigMap" (not (empty ($app.clusterValues).configMap) | ternary (printf "%s-cluster-values" (include "cluster.resource.name" $)) "")
        "clusterValuesSecret" (not (empty ($app.clusterValues).secret) | ternary (printf "%s-cluster-values" (include "cluster.resource.name" $)) "")
```

helm/cluster/values.schema.json
```diff
@@ -3677,7 +3677,7 @@
                 "kubernetesVersion": {
                     "type": "string",
                     "title": "Kubernetes version",
-                    "default": "1.25.16"
+                    "default": "1.33.4"
```
Gacko: Please use a more recent version here. We are introducing this in CAPI v35.0.0, which ships Kubernetes v1.35.x, so the oldest version we will face is Kubernetes v1.34.x during the upgrades, which is still in the version skew of v1.35.x. So a v1.35.x `kubectl` image should be fine. 🙂

Gacko:
In the last days I explored an alternative way for handling the handover from App CRs and Chart CRs to HelmRelease CRs. I talked to Piontec and Lukasz from Honeybadger about it and tested it thoroughly.

It first pauses all the Chart CRs, which would potentially trigger an `helm uninstall` either on the MC or the WC before executing the `helm upgrade` of the cluster chart. During the actual `helm upgrade` of the cluster chart, Helm calculates a diff between the old state and the desired state and finds App CRs to be deleted. It then deletes these App CRs and creates the new HelmRelease CRs.

Since the underlying Chart CRs are being paused, they are not getting deleted, but they already carry the deletion timestamp, so the moment they get unpaused and the finalizer gets removed, they immediately disappear without a `helm uninstall` getting triggered.

Chart CRs which are being created from App CRs installed by the customer will not get a deletion timestamp because their App CRs are not being managed by the cluster chart and for that reason the diff in `helm upgrade` does not delete them. So when we - in the end - remove the pause annotation and the finalizer, they just stay around and get picked up by Chart Operator again.

So make sure all App CRs are gone, also the ones owned by Observability Bundle or Security Bundle, I added a `kubectl wait` for HelmRelease CRs. This makes sure all HelmReleases are properly deployed and with that the underlying Helm releases are now being managed by HelmRelease CRs and Flux Helm Controller.

For this reason and because we as Tenet have to support and maintain this code in the end, also in case of incidents, we will go on with the new approach instead of the comprehensive shell script proposed earlier.

Gacko (approved):

PR #832, opened 2026-04-09
Add VPA CRD dependency for prometheus-blackbox-exporter App CR

T-Kukawka:
## Context

This PR is a follow-up to [giantswarm/giantswarm#35769](https://github.com/giantswarm/giantswarm/issues/35769) (adding VPA to phoenix components), specifically the `prometheus-blackbox-exporter` app.

[giantswarm/prometheus-blackbox-exporter-app#67](https://github.com/giantswarm/prometheus-blackbox-exporter-app/pull/67) adds a `VerticalPodAutoscaler` resource to the chart. This means the VPA CRDs must be present before that app is deployed — otherwise the helm installation fails with an unknown CRD error.

A parallel change was already made for `karpenter-taint-remover` in [giantswarm/cluster-aws#1858](https://github.com/giantswarm/cluster-aws/pull/1858), which follows the same pattern.

## Problem

`prometheus-blackbox-exporter` is an App CR (managed by app-operator) and already has one dependency:

```yaml
dependsOn: prometheus-operator-crd  # App CR
```

`vertical-pod-autoscaler-crd` is a **HelmRelease** (not an App CR), which requires the `app-operator.giantswarm.io/depends-on-helmrelease: "true"` annotation so app-operator knows to also look at HelmRelease objects when checking dependencies.

Additionally, the existing `dependsOn` field in the App CR file config only supported a **single string value**, making it impossible to express two dependencies without a template change.

## Investigation

Before implementing, we verified how app-operator handles the `depends-on-helmrelease` annotation by reading the source code (`service/controller/app/resource/chart/desired.go`):

- When `depends-on-helmrelease: "true"` is set, app-operator builds a **combined** map of both installed App CRs and HelmRelease objects.
- All dependencies listed in `app-operator.giantswarm.io/depends-on` are checked against this combined map — it is **additive**, not a switch.
- Setting the annotation does **not** drop App CR dependency checking.

This means we can safely add `vertical-pod-autoscaler-crd` (HelmRelease) alongside `prometheus-operator-crd` (App CR) in the same dependency list.

## Changes

- **`helm/cluster/templates/apps/apps.yaml`**: Extended the `dependsOn` handling to support both a string (existing behavior, backward-compatible) and a list. Uses `kindIs "string"` to detect which form is used.

- **`helm/cluster/files/apps/prometheus-blackbox-exporter.yaml`**: Changed `dependsOn` from a single string to a list with both dependencies, and added `dependsOnHelmRelease: true`.

## Test plan

- [ ] Verify that `prometheus-blackbox-exporter` App CR gets both `app-operator.giantswarm.io/depends-on` (comma-separated) and `app-operator.giantswarm.io/depends-on-helmrelease: "true"` annotations rendered correctly.
- [ ] Verify all other apps that use `dependsOn` as a string still render correctly (backward compatibility).
- [ ] Verify that on a new cluster, `prometheus-blackbox-exporter` waits for both `vertical-pod-autoscaler-crd` HelmRelease and `prometheus-operator-crd` App to be ready before installing.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


T-Kukawka:
## Compatibility analysis with app-operator

I verified this PR against the app-operator implementation in [`service/controller/app/resource/chart/desired.go`](https://github.com/giantswarm/app-operator/blob/main/service/controller/app/resource/chart/desired.go).

### How app-operator parses dependencies

```go
func getDependenciesFromCR(app v1alpha1.App) ([]string, error) {
    deps := make([]string, 0)
    dependsOn, found := app.Annotations[annotationChartOperatorDependsOn]
    if found {
        deps = strings.Split(dependsOn, ",")
    }
    // ...
}
```

app-operator reads the `app-operator.giantswarm.io/depends-on` annotation as a **single comma-separated string** and splits on `,`.

### What this PR renders

The template collects dependencies into a slice and joins with `,`:

```
app-operator.giantswarm.io/depends-on: {{ $dependencies | join "," | quote -}}
```

For `prometheus-blackbox-exporter` this produces:

```
app-operator.giantswarm.io/depends-on: "{cluster-name}-prometheus-operator-crd,{cluster-name}-vertical-pod-autoscaler-crd"
```

This is **exactly the format app-operator expects** — a comma-separated string. ✅

### HelmRelease dependency lookup

When `depends-on-helmrelease: "true"` is set, app-operator builds a combined lookup map by listing both App CRs and HelmReleases in the same namespace, keyed by `.GetName()`. It then checks all entries in the `depends-on` annotation against this combined map — so both `prometheus-operator-crd` (App CR) and `vertical-pod-autoscaler-crd` (HelmRelease) will be resolved correctly. ✅

### Conclusion

The implementation is correct and compatible with app-operator. The `kindIs "string"` guard for backward compatibility and the list-to-comma-join pattern match app-operator's expectations precisely.

AndiDog (commented):
Looks good. Let's remove draft status and E2E-test this.

AndiDog (approved):

njuettner:
With the migration what's been done by @fiunchinho here https://github.com/giantswarm/cluster/pull/780 does it still make sense to merge it, even we drop it App CR's soonish?

PR #851, opened 2026-04-30
Chart: Require `global.release.version` if using Releases to give a better rendering error message

AndiDog:
### What does this PR do?

To get a better error than

```
    Reason:         template: cluster-eks/charts/cluster/templates/clusterapi/workers/machinepool.yaml:18:8: executing "cluster-eks/charts/cluster/templates/clusterapi/workers/machinepool.yaml" at <include "cluster.labels.common" $>: error calling include: template: cluster-eks/charts/cluster/templates/_helpers.tpl:60:78: executing "cluster.labels.common" at <"v">: invalid value; expected string
    Status:         not-installed
```

Oh, and I'll fix kubectl-gs next – which didn't output the release version (for EKS).

Slightly related to my [EKS testing issue](https://github.com/giantswarm/giantswarm/issues/36119).

None, I think

No

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 - Control Plane: Make etcd image tag configurable. ([#841](https://github.com/giantswarm/cluster/pull/841))
 
+### Fixed
+
+- Chart: Require `global.release.version` if using Releases to give a better rendering error message.
+
 ## [6.4.0] - 2026-04-15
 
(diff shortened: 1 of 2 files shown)


Gacko (approved):
I wouldn't consider this a bug, so maybe just move your changelog entry to the `### Changed` section.

Apart from that: LGTM!

AndiDog:
@Gacko done. Should we run cluster test suites?

Gacko:
> @Gacko done. Should we run cluster test suites?

No, I don't think this is necessary.

PR #855, opened 2026-05-07
fix(selinux): cluster setup with selinux enforced

mirza-src:
### What does this PR do?
Towards https://github.com/giantswarm/giantswarm/issues/35494
Fixes issues in kubeadm script and kube-apiserver due to incorrect labels on the root filesystem that makes it impossible to start a cluster with SELinux enforced.

Clusters created using:
```yaml
global:
  components:
    selinux:
      mode: enforcing
    containerd:
      selinux:
        enabled: true
```
should now start but individual workload might still need additional policies to function properly.

No

### DIFF
--- a/helm/cluster/templates/clusterapi/controlplane/_helpers_files.tpl
+++ b/helm/cluster/templates/clusterapi/controlplane/_helpers_files.tpl
@@ -4,6 +4,7 @@
 {{- include "cluster.internal.controlPlane.kubeadm.files.audit" $ }}
 {{- include "cluster.internal.controlPlane.kubeadm.files.encryption" $ }}
 {{- include "cluster.internal.controlPlane.kubeadm.files.fairness" $ }}
+{{- include "cluster.internal.controlPlane.kubeadm.files.selinux" $ }}
 {{- include "cluster.internal.controlPlane.kubeadm.files.oidc" $ }}
 {{- include "cluster.internal.controlPlane.kubeadm.files.provider" $ }}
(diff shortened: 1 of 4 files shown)


stone-z (approved):

helm/cluster/templates/clusterapi/controlplane/_helpers_files.tpl
```diff
@@ -72,6 +73,15 @@
   content: {{ tpl ($.Files.Get "files/etc/kubernetes/patches/kube-controller-manager0+json.yaml") . | b64enc }}
 {{- end }}
 
+{{- define "cluster.internal.controlPlane.kubeadm.files.selinux" }}
+{{- if ne $.Values.global.components.selinux.mode "disabled" }}
```
Gacko, line 77: So empty string enables this? Is that intended? Else I'd rather compare to "enabled" here.

mirza-src: I assumed the values would be validated using the [`enum defined in the schema`](https://github.com/giantswarm/cluster/blob/e620ee73268413d7e6634da35015f66e84958697/helm/cluster/values.schema.json#L1689-L1699) and there is no `enabled` value for SELinux. If schema validation is not sufficient here, I can switch to an exhaustive check.

Gacko: kk, sorry, my fault!

Gacko (approved):

PR #858, opened 2026-05-08
Add "kyverno" as a dependency for the cluster-autoscaler App

iuriaranda:
### What does this PR do?

This is now needed since cluster-autoscaler deploys the PolicyException CRs in the `policy-exceptions` namespace, which is created by the kyverno App

### How does it look like?


iuriaranda:
this doesn't work. Dependencies come from the release

PR #887, opened 2026-05-28
refactor: Remove kamaji datastore/polex permissions handling

vxav:
Kamaji clusters require Datastore objects (etcd) which are cluster scoped.
The previous approach does not work in org namespaces other than `org-giantswarm` because of permissions.

I have moved the creation of the `clusterRole` to a global one in `Kamaji-app` (https://github.com/giantswarm/kamaji-app/pull/37) and the creation of `clusterRoleBindings` to org ns automation SAs to `rbac-operator` (https://github.com/giantswarm/rbac-operator/pull/724).


Gacko (approved):

glitchcrab (approved):
makes sense 🦀

vxav:
https://github.com/giantswarm/cluster/pull/887/commits/5a80e8fffa9f91ebfd720c74002e88c4aa0eadf0

> The test job in Run manifest tests started failing on this PR because PR #886 bumped actions/setup-go to v6, which defaults GOTOOLCHAIN=local. Without a Go version configured, setup-go uses the runner's preinstalled 1.24.13, and local blocks Go from auto-downloading the 1.25 toolchain required by helm/cluster/tests/go.mod. Setting go-version-file makes setup-go install the right version up front.

FYI @Gacko as you already approved.

vxav:
The original goal was for each app to own their own related components. It mades things too complex so that's out.

I simplified the thing, Polexes moved from kamaji-etcd chart to the cluster chart.
Removed the rbac and rbac cleanup nonsense which is no longer needed.

PR #892, opened 2026-06-01
feat(rbac): enable cluster-reader for default view bindings

pipo02mix:
Towards https://gigantic.slack.com/archives/C0ALXPMB1PW/p1780299573716949

- Bumps `rbac-bootstrap` chart `0.2.3` → `0.3.0`.
- Enables the new `clusterReader` toggle so the default `view` bindings for `giantswarm-admins` gain read access to nodes, PVs, CRDs, webhook configurations, and other cluster-scoped resources needed for troubleshooting workload clusters.

### DIFF
--- a/helm/cluster/files/helmreleases/rbac-bootstrap.yaml
+++ b/helm/cluster/files/helmreleases/rbac-bootstrap.yaml
@@ -4,9 +4,11 @@ configKey: rbacBootstrap
 namespace: kube-system
 # used by renovate
 # repo: giantswarm/rbac-bootstrap-app
-version: 0.2.3
+version: 0.3.0
 interval: 1m
 defaultValues:
+  clusterReader:
+    enabled: true
   bindings:
     - role: view
       groups:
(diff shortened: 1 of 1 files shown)


Gacko (changes requested):
Do not bump the version, please. First we have automation for it, second it is ignored anyway and only used for CI. 🙂

What I wanna tell: Yes, you can change it. But please do not expect it to have any effect somewhere outside CI. 😅

helm/cluster/files/helmreleases/rbac-bootstrap.yaml
```diff
@@ -4,9 +4,11 @@ configKey: rbacBootstrap
 namespace: kube-system
 # used by renovate
 # repo: giantswarm/rbac-bootstrap-app
-version: 0.2.3
+version: 0.3.0
```
pipo02mix:
```suggestion
version: 0.2.3
```

pipo02mix:
I have reverted it. I hope default values are taking into account :)

Gacko (approved):

PR #900, opened 2026-06-05
feat: add support for AKS clusters

iuriaranda:
### What does this PR do?

In AKS, cilium, coredns and metrics-server are built-in and we don't deploy our own apps, so dependant apps don't need to specify the dependency.

Towards https://github.com/giantswarm/giantswarm/issues/36639

### How does it look like?

### DIFF
--- a/helm/cluster/files/apps/observability-bundle.yaml
+++ b/helm/cluster/files/apps/observability-bundle.yaml
@@ -6,7 +6,7 @@ clusterValues:
   secret: false
 inCluster: true
 dependsOnHelmRelease: true
-dependsOn: coredns
+dependsOn: {{ if not (eq $.Values.providerIntegration.provider "aks") }}coredns{{ else }}[]{{ end }}
 namespace: kube-system
 # used by renovate
 # repo: giantswarm/observability-bundle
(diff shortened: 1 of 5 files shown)


helm/cluster/files/apps/observability-bundle.yaml
```diff
@@ -6,7 +6,7 @@ clusterValues:
   secret: false
 inCluster: true
 dependsOnHelmRelease: true
-dependsOn: coredns
+dependsOn: {{ if not (eq $.Values.providerIntegration.provider "aks") }}coredns{{ else }}[]{{ end }}
```
Gacko: You're mixing `string` and `array` here. Can you check the type of `dependsOn` and make it right? Thank you!

Gacko (changes requested):

helm/cluster/files/apps/observability-bundle.yaml
iuriaranda: yeah, just saw. Apparently this form works for apps (was already like that for cluster-autoscaler), but it doesn't in helmreleases. I've fixed it everywhere

iuriaranda:
@Gacko I've fixed the array format. Any idea why the documentation checks are failing?

Gacko (approved):
LGTM! In general dependencies are mostly defined by the Release CR in the end, but it's never a bad idea to have it reflected here.

Gacko:
> Any idea why the documentation checks are failing?

The Go binary got updated (new release), but the GitHub action running it does not ship an up-to-date Go version.

PR #901, opened 2026-06-05
feat: add support for disabling the external autoscaler annotation on MachinePools

iuriaranda:
### What does this PR do?

Needed on AKS clusters, where autoscaling is done by AKS and CAPZ doesn't expect to have such an annotation set.

Defaults to existing behavior.

Towards https://github.com/giantswarm/giantswarm/issues/36639

### How does it look like?


Gacko (changes requested):
Please do it in the form of `externalAutoscaler.enabled`. Also since this is for the MachinePool, wouldn't it make more sense to refactor `machinePoolResourcesEnabled` to an object named `machinePoolResources` with sub-properties?

iuriaranda:
> Please do it in the form of `externalAutoscaler.enabled`. Also since this is for the MachinePool, wouldn't it make more sense to refactor `machinePoolResourcesEnabled` to an object named `machinePoolResources` with sub-properties?

@Gacko I've refactored it in the form of `externalAutoscaler.enabled` as you suggested.

As per the `machinePoolResourcesEnabled` suggestion, I think it would only make things more confusing. According to the [readme](https://github.com/giantswarm/cluster/blob/main/README.md#node-pool-resources), that flag controls whether the node pool resources are rendered or not (not only `MachinePools`), and the name is somehow misplaced.

Gacko:
But isn't your change mostly affecting the machine pool? 😅 At least I do not see some kind of external autoscaler resource in the chart which could get configured via this property.

From my understanding, the whole resources API is made for controlling the resource types the cluster chart may render. Rocket for example added support to deploy a different control plane resource in the `controlPlane` object.

So if your change is meant to configure the machine pool or deployment in a way, then I'd assume the best place for these properties is the `machinePoolResources` property.

Gacko (approved):

PR #909, opened 2026-06-10
feat: honor extraConfigs priority in HelmRelease valuesFrom ordering

fiunchinho:
Towards https://github.com/giantswarm/giantswarm/issues/36096

Stacked on #780.

## Problem

App CRs support a `priority` field (1-150, default 25) on `extraConfigs` entries that fixes each entry's merge position relative to the platform config layers — catalog (slot 0), cluster config (slot 50), user config (slot 100) — regardless of position in the list (see the [multi-layer-app-config RFC](https://github.com/giantswarm/rfc/tree/main/multi-layer-app-config)). Flux merges `valuesFrom` strictly in list order and merges inline `spec.values` last, so the migrated HelmReleases silently ignore priorities. Bundle customers are known to rely on them.

## Changes

1. **Stable priority sort + slot interleaving.** `extraConfigs` are bucketed with a counting sort: entries with priority ≤ 50 render before the `-cluster-values` references, 51-100 after them, and > 100 last. Ties keep their authored order (matching App platform behavior).

2. **User values via ConfigMap when the > 100 band is used.** Flux merges inline `spec.values` after every `valuesFrom` entry, so an extraConfig with priority > 100 (meant to override user config) could never win against inline values. When such an entry exists for an app, the merged default+provider+user values are delivered through a `<cluster>-<app>-user-values` ConfigMap referenced at the user-config slot, and the > 100 entries follow it. Apps without > 100 entries keep the inline form — for them, inline-last is exactly equivalent to the user-config slot.

The CI values already exercise the > 100 band (cert-manager 110/120, external-dns 120), so `helm template` diffs demonstrate the fix directly.

## Verification

- Baseline render against `ci/ci-values.yaml`: only cert-manager and external-dns change (the two apps with priority > 100 entries) — they switch from inline values to the user-values ConfigMap, ordered `user-values → 110 → 120`.
- Three-band scenario (priorities 10/120/75 authored in non-priority order): renders `10 → 75 → user-values → 120`.
- All `ci/*.yaml` values files render cleanly; `make test` passes.

## Follow-ups (not in this PR)

- Same treatment for the provider-chart HR templates (cluster-aws) and the bundle charts (security-bundle, observability-bundle, aws-nth-bundle) — the issue calls out bundle customers specifically.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,6 +15,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 - Migrate default applications from App CRs to Flux HelmRelease CRs.
 - Add support for AKS clusters
+- HelmReleases: honor the App platform `priority` field (1-150, default 25) on `extraConfigs` entries. `valuesFrom` entries are now ordered by priority around the cluster config (slot 50) and user config (slot 100) layers instead of list order, preserving the App CR merge semantics after the migration. When an extraConfig has priority above 100, the app's merged values are delivered via a `<cluster>-<app>-user-values` ConfigMap at the user-config slot instead of inline `spec.values`, so that entry can still override user config. ([giantswarm#36096](https://github.com/giantswarm/giantswarm/issues/36096))
 
 ### Fixed
 
--- a/helm/cluster/ci/test-extraconfigs-priority-ordering-values.yaml
+++ b/helm/cluster/ci/test-extraconfigs-priority-ordering-values.yaml
@@ -58,10 +58,7 @@ global:
         - kind: secret
           name: sec-pre-cluster-1
           priority: 30
-          # sortAlpha is case-sensitive, a capital letter is sorted before the lowercase one.
-          # The value schema already prevents that, but not all systems might run the checks or can be skipped too.
-          # The sort helper converts them to a numerical value and calls lower() too.
-        - kind: CONFIGMAP
+        - kind: configMap
           name: cm-pre-cluster
           priority: 20
         - kind: secret
@@ -75,7 +72,7 @@ global:
         - kind: configMap
           name: cm-pre-user
           priority: 90
-        - kind: SECRET
+        - kind: secret
           name: sec-pre-cluster-2
           priority: 30
         - kind: configMap
@@ -93,10 +90,7 @@ global:
         - kind: secret
           name: sec-pre-cluster-1
           priority: 30
-          # sortAlpha is case-sensitive, a capital letter is sorted before the lowercase one.
-          # The value schema already prevents that, but not all systems might run the checks or can be skipped too.
-          # The sort helper converts them to a numerical value and calls lower() too.
-        - kind: CONFIGMAP
+        - kind: configMap
           name: cm-pre-cluster
           priority: 20
         - kind: secret
@@ -104,7 +98,7 @@ global:
           priority: 130
         - kind: configMap
           name: cm-default        # no priority -> defaults to 25
-        - kind: SECRET
+        - kind: secret
           name: sec-pre-user
           priority: 90
         - kind: configMap
--- a/helm/cluster/ci/test-extraconfigs-priority-ordering-values.yaml
+++ b/helm/cluster/ci/test-extraconfigs-priority-ordering-values.yaml
@@ -1,0 +1,142 @@
+# Exercises the App-platform `priority` ordering of extraConfigs as reproduced
+# in the HelmRelease `spec.valuesFrom` list (see the `cluster.app.sortedValuesFrom`
+# helper and giantswarm/giantswarm#36096).
+#
+# App platform built the configMap layers and the secret layers in two independent
+# passes and merged the secret blob OVER the configMap blob, so:
+#   * ALL configMaps merge before ALL secrets (a secret always overrides a configMap,
+#     regardless of priority);
+#   * within each kind the order is, by priority band:
+#       extras priority <= 50  ->  cluster-values slot (50)  ->
+#       extras 50 < priority <= 100  ->  user-values slot (100)  ->
+#       extras priority > 100
+#   * equal-priority extras keep their authored order;
+#   * an extra with no priority defaults to 25.
+#
+# Flux merges `valuesFrom` strictly top-to-bottom (later wins), so the rendered list
+# must already be in that order. The merged default+provider+user values are delivered
+# through a `<cluster>-<app>-user-values` ConfigMap at the user-config slot (NOT inline
+# `spec.values`, which Flux always merges last).
+#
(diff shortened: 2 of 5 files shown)


mproffitt:
I was reading through the logic of this and one thing immediately jumped out is that in its present form, it does not seem to honour the priority within each individual bucket, but simply appends into each bucket depending on whether or not the priority is less than the ceiling of that bucket.

This would lead to a scenario whereby customer configs are given to the helmrelease out of order, potentially causing some expected values to be overridden incorrectly.

I spoke with both @uvegla and @ljakimczuk about this and whilst the logic is more complex, we propose something more akin to the following (courtesy of LukaszJ)

```
{{- define "foo.sortedExtraConfigs" -}}
{{- $keyed := dict -}}
{{- $keys := list -}}

{{- if .Values.config.configMap -}}
{{- $key := printf "%s_%010d_%s_%s_%010d" "0" 50 "0" .Values.config.configMap.name 0 -}}
{{- $keyed = set $keyed $key (dict "kind" "configmap" "name" .Values.config.configMap.name) -}}
{{- $keys = append $keys $key -}}
{{- end -}}

{{- if .Values.config.secret -}}
{{- $key := printf "%s_%010d_%s_%s_%010d" "1" 50 "0" .Values.config.secret.name 0 -}}
{{- $keyed = set $keyed $key (dict "kind" "secret" "name" .Values.config.secret.name) -}}
{{- $keys = append $keys $key -}}
{{- end -}}

{{- if .Values.userConfig.configMap -}}
{{- $key := printf "%s_%010d_%s_%s_%010d" "0" 100 "0" .Values.userConfig.configMap.name 0 -}}
{{- $keyed = set $keyed $key (dict "kind" "configmap" "name" .Values.userConfig.configMap.name) -}}
{{- $keys = append $keys $key -}}
{{- end -}}

{{- if .Values.userConfig.secret -}}
{{- $key := printf "%s_%010d_%s_%s_%010d" "1" 100 "0" .Values.userConfig.secret.name 0 -}}
{{- $keyed = set $keyed $key (dict "kind" "secret" "name" .Values.userConfig.secret.name) -}}
{{- $keys = append $keys $key -}}
{{- end -}}

{{- range $index, $config := .Values.extraConfigs -}}
{{- $kindOrder := "1" -}}
{{- if eq (lower $config.kind) "configmap" -}}
{{- $kindOrder = "0" -}}
{{- end -}}
{{- $priority := printf "%010d" (int (default 25 $config.priority)) -}}
{{- $key := printf "%s_%s_%s_%s_%010d" $kindOrder $priority "1" $config.name $index -}}
{{- $keyed = set $keyed $key $config -}}
{{- $keys = append $keys $key -}}
{{- end -}}
{{- $sorted := list -}}
{{- range $key := sortAlpha $keys -}}
{{- $sorted = append $sorted (get $keyed $key) -}}
{{- end -}}
{{- $sorted | toYaml -}}
{{- end }}
```

This skips the buckets entirely, instead relying on a keyed map and a sorted list of map keys using the `sortAlpha` function to sort the list based on its priority stored as part of the key name.

This means that for the example:

<details>
<summary>AppCR configs</summary>

```
config:
  configMap:
    name: foo-values
  secret:
    name: foo-secrets
extraConfigs:
  - kind: configmap
    name: config-to-apply-before-foo-values
    priority: 10
  - kind: secret
    name: my-configuration-b
    priority: 25
  - kind: secret
    name: my-configuration-a
    priority: 25
  - kind: configMap
    name: low-priority-config
    priority: 100
  - kind: configMap
    name: foo-values-level-config
    priority: 25
  - kind: configMap
    name: high-priority-config
    priority: 5
  - kind: secret
    name: secreet-for-before-config
    priority: 20
  - kind: secret
    name: secreet-user-config-level
    priority: 100
userConfig:
  configMap:
    name: foo-user-values
  secret:
    name: foo-user-secrets
```

</details>

the helmrelease would be rendered as:

<details>
<summary>Final HelmRelease</summary>

```
apiVersion: helm.toolkit.fluxcd.io/v2
kind: HelmRelease
metadata:
  name: podinfo
  namespace: default
spec:
  interval: 10m
  chartRef:
    kind: HelmChart
    name: podinfo
    namespace: default
  valuesFrom:
  - kind: ConfigMap
    name: high-priority-config
  - kind: Configmap
    name: config-to-apply-before-foo-values
  - kind: Configmap
    name: foo-values
  - kind: ConfigMap
    name: foo-values-level-config
  - kind: Configmap
    name: foo-user-values
  - kind: ConfigMap
    name: low-priority-config
  - kind: Secret
    name: secreet-for-before-config
  - kind: Secret
    name: foo-secrets
  - kind: Secret
    name: my-configuration-a
  - kind: Secret
    name: my-configuration-b
  - kind: Secret
    name: foo-user-secrets
  - kind: Secret
    name: secreet-user-config-level
```

</details>

This is effectively the same ordering as we apply in AppOp/ChartOp combo with both configmaps and secrets maintaining their respective priorities, then secrets having priority over CM in the final rendered output.

Use the function like this:

<details>
<summary>Usage</summary>

```
apiVersion: helm.toolkit.fluxcd.io/v2
kind: HelmRelease
metadata:
  name: podinfo
  namespace: default
spec:
  interval: 10m
  chartRef:
    kind: HelmChart
    name: podinfo
    namespace: default
  valuesFrom:
  {{- range include "foo.sortedExtraConfigs" . | fromYamlArray }}
  - kind: {{ .kind | title }}
    name: {{ .name }}
  {{- end }}
```

</details>

fiunchinho:
I've tested this again, and it actually does sort within each bucket (the outer loop is a counting sort over `untilStep 1 151 1`), e.g. authored 40,10,30 renders 10,30,40. But you made me realize about a different issue: my implementation interleaves configmaps and secrets by priority, but in the app platform a secret always overrides a configMap, regardless of priority.

I updated the PR to solve this issue based on your proposed solution (thank you for the snippet, it's most of the way there). Apart from fixing the Secret vs ConfigMap priority, another big change is that user values move into the `<cluster>-<app>-user-values` ConfigMap at the user-config slot, instead of inline `spec.values`. Flux always merges `spec.values` last, so anything in the `>100` band (or any secret) could never override inline user values. This is also what the App CRs did, so it's not new behaviour.

So now this PR contains a whole new approach based on your proposed solution. To make this easy to discuss and to guard against regressions, I added the helm values file `ci/test-extraconfigs-priority-ordering-values.yaml`. It contains configMap/secret extraConfigs authored deliberately out of order, on both a WC-targeted app and an MC-targeted bundle, with the expected order documented in the header. You can render it directly with `helm template mycluster helm/cluster -n org-giantswarm --values helm/cluster/ci/test-extraconfigs-priority-ordering-values.yaml`.

uvegla (approved):

PR #930, opened 2026-06-22
chore: migrate cert-manager values for v4.0.1 release

fhielpos:
> [!WARNING]
> This PR depends on the v4.0.1 cert-manager release:
> https://github.com/giantswarm/cert-manager-app/releases/tag/v4.0.1

`cert-manager` now vendors upstream `cert-manager` as a subchart, so its upstream values live under a `cert-manager`: key. The values we pass to `cert-manager-app` from the cluster charts need to move accordingly.

Towards: https://github.com/giantswarm/giantswarm/issues/36890

### DIFF
--- a/helm/cluster/files/apps/cert-manager.yaml
+++ b/helm/cluster/files/apps/cert-manager.yaml
@@ -10,13 +10,14 @@ namespace: kube-system
 # repo: giantswarm/cert-manager-app
 version: 4.0.1
 defaultValues:
-  # cert-manager's DNS01 solver by default tries to reach authoritative nameservers directly, using
-  # their public IPs. Since those are not reachable from private clusters, we instead rely on the
-  # recursive nameserver (e.g. for AWS, that's normally the default nameserver requested in EC2 instances).
-  dns01RecursiveNameserversOnly: true
-  ingressShim:
-    defaultIssuerName: letsencrypt-giantswarm
-    defaultIssuerKind: ClusterIssuer
-    defaultIssuerGroup: cert-manager.io
+  cert-manager:
+    # cert-manager's DNS01 solver by default tries to reach authoritative nameservers directly, using
+    # their public IPs. Since those are not reachable from private clusters, we instead rely on the
+    # recursive nameserver (e.g. for AWS, that's normally the default nameserver requested in EC2 instances).
+    dns01RecursiveNameserversOnly: true
+    ingressShim:
+      defaultIssuerName: letsencrypt-giantswarm
+      defaultIssuerKind: ClusterIssuer
+      defaultIssuerGroup: cert-manager.io
   ciliumNetworkPolicy:
     enabled: true
--- a/helm/cluster/files/apps/cert-manager.yaml
+++ b/helm/cluster/files/apps/cert-manager.yaml
@@ -12,8 +12,11 @@ version: 4.0.1
 defaultValues:
   # Since v4.0.0, cert-manager upstream values need to be under the cert-manager key.
   # This gates the config based on the cert-manager app version
+  {{- $certManagerVersion := "N/A" }}
+  {{- if $.Values.providerIntegration.useReleases }}
   {{- $_ := set $ "appName" "cert-manager" }}
-  {{- $certManagerVersion := include "cluster.app.version" $ | trim }}
+  {{- $certManagerVersion = include "cluster.app.version" $ | trim }}
+  {{- end }}
   {{- if or (eq $certManagerVersion "N/A") (semverCompare ">=4.0.0" $certManagerVersion) }}
   cert-manager:
     # cert-manager's DNS01 solver by default tries to reach authoritative nameservers directly, using
--- a/helm/cluster/files/apps/cert-manager.yaml
+++ b/helm/cluster/files/apps/cert-manager.yaml
@@ -10,6 +10,14 @@ namespace: kube-system
 # repo: giantswarm/cert-manager-app
 version: 4.0.1
 defaultValues:
+  # Since v4.0.0, cert-manager upstream values need to be under the cert-manager key.
+  # This gates the config based on the cert-manager app version
+  {{- $certManagerVersion := "N/A" }}
+  {{- if $.Values.providerIntegration.useReleases }}
+  {{- $_ := set $ "appName" "cert-manager" }}
+  {{- $certManagerVersion = include "cluster.app.version" $ | trim }}
+  {{- end }}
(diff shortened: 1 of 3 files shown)


Gacko (changes requested):
It seems your branch is somehow heavily outdated. Please rebase it on top of `main` and run `make generate`.

Gacko (changes requested):
I just double-checked and this moved the proxy configuration values to `cert-manager`, too, while the cluster values still have it on root level. Can you please make sure all the values you're consuming from the cluster values continue to get provided to the new sub-property?

fhielpos:
Updated `cert-manager` to handle `.Values.proxy` as well as `.Values.cluster.proxy` here:
https://github.com/giantswarm/cert-manager-app/pull/640
https://github.com/giantswarm/cert-manager-app/pull/639

Gacko (approved):

Gacko (changes requested):
If you are afraid of issues due to using this code in conjunction with `cert-manager` < v4.0.0, then please have a look here: https://github.com/giantswarm/cluster-eks/blob/main/helm/cluster-eks/templates/apps/_teleportkubeagent_app_config.yaml.

Because if we merge, release and integrate this into `cluster-<provider>` charts, any kind of cluster test suites will start to fail as they use the latest available CAPI Release which currently does not include `cert-manager` v4.0.0.

Gacko:
And while you're on it, it would be cool if you could also implement the required changes here: https://github.com/giantswarm/mc-bootstrap/pull/1664.

fhielpos:
@Gacko If this looks good to you I will replicate it to the other repos 😄

helm/cluster/files/apps/cert-manager.yaml
```diff
@@ -10,13 +10,30 @@ namespace: kube-system
...
-  # their public IPs. Since those are not reachable from private clusters, we instead rely on the
-  # recursive nameserver (e.g. for AWS, that's normally the default nameserver requested in EC2 instances).
+  # Since v4.0.0, cert-manager upstream values need to be under the cert-manager key.
+  # This gates the config based on the cert-manager app version
+  {{- $certManagerVersion := "N/A" }}
+  {{- if $.Values.providerIntegration.useReleases }}
+  {{- $_ := set $ "appName" "cert-manager" }}
+  {{- $certManagerVersion = include "cluster.app.version" $ | trim }}
+  {{- end }}
+  {{- if or (eq $certManagerVersion "N/A") (semverCompare ">=4.0.0" $certManagerVersion) }}
```
Gacko, line 20: There's no need to check `providerIntegration.useReleases` or set the version to `N/A`. `cluster.app.version` emits `N/A` if there's no release or the app is missing. 🙂 

https://github.com/giantswarm/cluster/blob/main/helm/cluster/templates/_helpers.tpl#L267-L288

Gacko (changes requested):

helm/cluster/files/apps/cert-manager.yaml
fhielpos: CI was complaining before, let me remove it and check again

fhielpos: I removed it and it failed 😅 
```
[ERROR] templates/: template: cluster/templates/apps/apps.yaml:9:12: executing "cluster/templates/apps/apps.yaml" at <tpl (.Files.Get $path) $>: error calling tpl: error during tpl function execution for "appName: cert-manager\nchartName: cert-manager-app\nconfigKey: certManager\nclusterValues:\n  configMap: true\n  secret: true\ndependsOn: prometheus-operator-crd\nnamespace: kube-system\n# used by renovate\n# repo: giantswarm/cert-manager-app\nversion: 4.0.1\ndefaultValues:\n  # Since v4.0.0, cert-manager upstream values need to be under the cert-manager key.\n  # This gates the config based on the cert-manager app version\n  {{- $_ := set $ \"appName\" \"cert-manager\" }}\n  {{- $certManagerVersion := include \"cluster.app.version\" $ | trim }}\n  {{- if or (eq $certManagerVersion \"N/A\") (semverCompare \">=4.0.0\" $certManagerVersion) }}\n  cert-manager:\n    # cert-manager's DNS01 solver by default tries to reach authoritative nameservers directly, using\n    # their public IPs. Since those are not reachable from private clusters, we instead rely on the\n    # recursive nameserver (e.g. for AWS, that's normally the default nameserver requested in EC2 instances).\n    dns01RecursiveNameserversOnly: true\n    ingressShim:\n      defaultIssuerName: letsencrypt-giantswarm\n      defaultIssuerKind: ClusterIssuer\n      defaultIssuerGroup: cert-manager.io\n  {{- else }}\n  # Duplicated block for < v4.0.0 versions\n  dns01RecursiveNameserversOnly: true\n  ingressShim:\n    defaultIssuerName: letsencrypt-giantswarm\n    defaultIssuerKind: ClusterIssuer\n    defaultIssuerGroup: cert-manager.io\n  {{- end }}\n  ciliumNetworkPolicy:\n    enabled: true\n": template: gotpl:16:29: executing "gotpl" at <include "cluster.app.version" $>: error calling include: template: cluster/templates/_helpers.tpl:279:11: executing "cluster.app.version" at <include "cluster.internal.get-release-resource" $>: error calling include: template: cluster/templates/_helpers.tpl:257:68: executing "cluster.internal.get-release-resource" at <"v">: invalid value; expected string
```

Gacko: I'm gonna have a look at this later. Because IMHO it should work like this and we should rather fix the root cause than the symptom.

Gacko:
As suspected: Not your fault. See https://github.com/giantswarm/cluster/pull/938/changes/cd03b922de191a816e315ace59b82ee00ef352f7 for the fix.

Gacko:
Ok, nvm, it just fails elsewhere then.

helm/cluster/templates/_helpers.tpl
```diff
@@ -276,6 +276,7 @@ Where `data` is the data to hash and `global` is the top level scope.
 */}}
 {{- define "cluster.app.version" }}
 {{- $appVersion := "N/A" }}
+{{- if $.Values.providerIntegration.useReleases }}
```
Gacko: Please don't do that. 😅 I think having it where you're using it is fine.

fhielpos: Alright, let's do that then 😄

Gacko (approved):

PR #944, opened 2026-07-02
feat: Support templating on the `global.apps.<name>.extraConfigs.name` field

robinkb:
### What does this PR do?

Adds support for templating the `global.apps.<name>.extraConfigs.name` field.

Allows loading dynamically named ConfigMaps as extraConfigs.
This is handy for ConfigMaps that are generated on the MC that contain dynamic values that must be injected into an App or HelmRelease.

### How does it look like?

Provider charts can do this:

```yaml
global:
  apps:
    externalDns:
      extraConfigs:
        - kind: configMap
          name: "{{ $.Values.global.metadata.name }}-external-dns-identity"
```

Related to https://github.com/giantswarm/giantswarm/issues/36640.

The schema description has been updated and docs regenerated.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Changed
+
+- Support templating on the `global.apps.<name>.extraConfigs.name` field.
+
 ## [6.7.0] - 2026-06-17
 
 ### Changed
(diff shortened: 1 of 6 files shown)


Gacko (changes requested):
Please have a look at this: https://github.com/giantswarm/cluster/pull/780/changes#diff-ca2e24f676b1504a14497a78e835bd9c9d404d2edebf58eda2c832cbbf2421d9R616.

We are already implementing this there and it will get shipped in CAPI v35.0.0. 🙂 So if your change isn't urgent, I'd ask you to wait until we merged and released the above PR. WDYT?

robinkb:
I need this for a PoC that is kicking off this month, but I could use the build from this PR instead until #780 gets merged.

Gacko (changes requested):
Alright, whatever you prefer. I mean, we could theoretically get this merged and released, but since we are working on CAPI v35.0.0 anyway already, I'd prefer not doing the same work (releasing `cluster`, integrating into `cluster-<provider>`,  releasing `cluster-<provider>`, creating a CAPI Release) twice. 😅

Gacko (approved):
My fault, wanted to approve. Sorry!

robinkb:
I'll move this to Draft and wait for #780 to be completed instead :)

This will be enough until then. No need to do more release work and merge conflict resolution than necessary.

robinkb:
In the end, we decided to push this through and not wait for #780. We cannot really wait for that PR to be finished, and given the small size of this PR, we don't think that it will cause big (if any) merge conflicts.

PR #960, opened 2026-07-22
feat: opt-in local instance-store NVMe ephemeral storage for workers

luca-rui:
## What

Adds a supported, opt-in per-node-pool feature for using local EC2 instance-store NVMe disks as worker ephemeral storage, replacing the unsupported `internal.advancedConfiguration.workers.preKubeadmCommands` escape hatch currently used for this.

When **any** node pool sets `localNvme.enabled: true` (schema for that flag lives in the provider chart, e.g. `cluster-aws`), a self-gating bootstrap script is rendered once into the worker `preKubeadmCommands`. At runtime it:

1. Detects instance-store NVMe by the Nitro controller model string (`Amazon EC2 NVMe Instance Storage`), RAID0-stripes them if there is more than one, formats XFS, and mounts over the kubelet directory (default `/var/lib/kubelet`) with a `nofail` fstab entry — so pod `emptyDir` volumes land on the fast node-local NVMe instead of the EBS lib/root volume.
2. Writes a `+merge` kubelet patch adding `evictionHard.nodefs.available` back (GS's `evictionHard` otherwise replaces kubelet's built-in `nodefs.available<10%` default, leaving no disk backstop).

The script is a **safe no-op** on instance types without instance store, which is why it can be rendered cluster-wide.

## How it's wired

- `cluster.internal.workers.localEphemeralStorage.enabled` aggregates over `global.nodePools` (mirrors the existing `hasKarpenterNodePool`/`hasAWSMachinePools` helpers). The aggregation is done inside this chart because Helm can't have a parent chart write computed values into a subchart.
- Cluster-wide parameters (`mountPath`, `evictionHardNodefsAvailable`) live under `internal.advancedConfiguration.workers.localEphemeralStorage`.

## Testing

`helm template` with the opt-in on injects the command into worker configs; off injects nothing; the rendered base64 decodes back to the script with parameters substituted. `helm lint` clean.

## Note

No `Chart.yaml` bump — version is cut by release automation. Only a `## [Unreleased]` CHANGELOG entry is added here. The `cluster-aws` counterpart (per-pool `localNvme.enabled` schema) is a separate PR, gated on this being released.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


Gacko (changes requested):
This reads like an AWS specific, so please do not put this into the `cluster` chart but into `cluster-aws`. 🙂

Gacko (commented):
Sorry, I missed that line:

> The aggregation is done inside this chart because Helm can't have a parent chart write computed values into a subchart.

Gonna check later.

Gacko (changes requested):
Please do not put provider specific logic and/or terms into this chart.

You can always make `cluster` chart call a helper defined in the provider specific chart if there is something you need to hand in dynamically. This is how we provide provider specific app configuration to default apps rendered in the `cluster` chart.

Also you should not use `echo` to create the script file. There are ways to add files natively.

Last but not least I'd appreciate to have an issue on our board so an engineer can have a look at how to implement this in a suitable way.

luca-rui:
Thanks for the review, @Gacko, agreed on all three points. I reworked the PR from scratch on top of `main`:

- The AWS-specific code is gone from this chart. `cluster` now only adds a hook,
  `providerIntegration.workers.kubeadmConfig.preKubeadmCommandsTemplateName`. It works like the other
  `*TemplateName` hooks: the provider chart names one of its templates, and this chart includes it in the worker
  `preKubeadmCommands` once per node pool with the root context (`$.nodePool.name`, `$.nodePool.config`), so the
  provider can emit commands for selected node pools only.
- No new customer-facing values here. `localNvme` moves to giantswarm/cluster-aws#2039 together with the script,
  which ships through the provider-specific-files Secret and `kubeadmConfig.files` instead of an inline base64
  command.
- Added `ci/test-prekubeadmcommands-templatename-values.yaml` and a ginkgo test: two node pools, each
  `KubeadmConfig` gets its own command.

The board issue for the follow-ups is giantswarm/giantswarm#37692. Once this is released, giantswarm/cluster-aws#2039 bumps the dependency and leaves draft.

Gacko (changes requested):
btw, I just checked and there is a node pool available in the context of these helpers, so you could run your script for specific node pools only.

See how you did it in `cluster.test.workers.kubeadm.preKubeadmCommands.provider`.

Gacko (changes requested):
Please update the PR description and title so it reflects what you're actually changing independent of why you're doing it. You can still have some context/examples, but it should be clear what people can do with your feature.

helm/cluster/templates/clusterapi/workers/_helpers_prekubeadmcommands.tpl
```diff
@@ -13,11 +13,23 @@
...
+      rendered once per node pool with the root context, so it can use $.nodePool.name and
+      $.nodePool.config to emit commands for specific node pools only.
+*/}}
 {{- define "cluster.internal.workers.kubeadm.preKubeadmCommands.provider" }}
 {{- range $command := $.Values.providerIntegration.workers.kubeadmConfig.preKubeadmCommands }}
 - {{ $command }}
 {{- end }}
+{{- with $.Values.providerIntegration.workers.kubeadmConfig.preKubeadmCommandsTemplateName }}
+{{- include . $ }}
+{{- end }}
```
Gacko: I know, `with` is shiny on objects, but definitely not easy to read and understand on simple strings. 😃

Please use a simple `if` here and name the variable once again when using it.

Gacko (changes requested):
One last thing: Can you also implement this for control plane nodes and post-kubeadm commands? That would be cool!

helm/cluster/templates/clusterapi/workers/_helpers_prekubeadmcommands.tpl
luca-rui: Done in 80844d1.

luca-rui:
Thanks @Gacko, all four points are addressed. Two new commits on top of 9c05e20 so the delta is easy to see:

- `with` -> `if`, value named again: 80844d1.
- Control plane and post-kubeadm: 5ff9b52 adds `providerIntegration.controlPlane.kubeadmConfig.preKubeadmCommandsTemplateName`, `providerIntegration.controlPlane.kubeadmConfig.postKubeadmCommandsTemplateName` and `providerIntegration.workers.kubeadmConfig.postKubeadmCommandsTemplateName`, same idiom as the existing workers pre-kubeadm hook. The control plane templates are rendered once with the root context and must not read `$.nodePool` (noted in the helper comments). The ci values file and the ginkgo test now cover all four hooks and check that no command leaks between phases, roles or node pools.
- Node pool in context: yes, that is what the consumer does. giantswarm/cluster-aws#2039 sets the workers pre-kubeadm hook to `awsWorkersPreKubeadmCommands`, which emits the setup script only when `$.nodePool.config.localNvme.enabled` is true. The old PR description still said "cluster-wide", sorry for the confusion.
- Title and description rewritten to describe the four hooks; the customer story moved to the background section.

On the helm-diff job: it cannot diff the new ci values file against the base (`Diff error`), and the `managed-by` label change on the kamaji `PolicyException`s is older than this branch (`main` and v8.1.0 both carry `kyverno-policy-operator`). Neither comes from these commits. Ready for another look.

Gacko (changes requested):
Basically LGTM, but maybe make clear the result of that template has to be a YAML array of strings and also make sure your code tries to read the returned value as such using `fromYamlArray` and then iterates over them to render them as the already existing static values do.

I'm thinking of something like:

```
{{- range $command := include $.Values.providerIntegration.controlPlane.kubeadmConfig.postKubeadmCommandsTemplateName $ | fromYamlArray }}
- {{ $command }}
{{- end }}
```

luca-rui:
Thanks @Gacko, done in a56309e. The four `.provider` helpers now read the template output with `fromYamlArray` and render every item like the static commands, same as the `connectivity.proxy.noProxy.templateName` hook. The contract (the named template must render a YAML list of strings, one command per item) is stated in the helper comments, the schema descriptions/README and the CHANGELOG bullet; the PR description says the same. I kept the `if` guard around the `range` because `include` with an empty name fails.

Checked locally: the ci values file renders byte-identical before and after the change; a template that renders nothing produces no `postKubeadmCommands` entry (`fromYamlArray` of an empty string is an empty list); output that is not a list ends up as a single item carrying the parse error text, the same behaviour as the noProxy hook; the other 16 ci values files still render identically to `main`; `make generate` is clean and `make test` passes. 707c22a only merges `main` (two dependency bumps) so the branch is up to date for the required checks. Ready for another look.

Gacko (approved):

Gacko:
Not here, on the `cluster-aws` PR. 🙃 Make sure to reference the development version of this chart here in the `Chart.yaml` there.

